PR #3, opened 2020-06-16
[13.0][ADD] payroll_account copy from Odoo12

Saran440:
- Change name from `hr_payroll_account` to `payroll_account`
- Copy from Odoo12 and migration to version 13.0
- Depend on payroll


pedrobaeza:
Closing in favor of #4

PR #7, opened 2020-10-08
hr_payroll: fix contract views

davejames:
Previously the extra fields added by this module were positioned relative to the wage and resource_calendar_id fields, however these fields are inside div tags which resulted in the new fields being irregularly placed and missing their labels

**Before:**
image
image

**After:**
image
image


davejames:
closing in favour of #8

PR #5, opened 2020-09-01
[13.0][MIG] payroll_account from odoo v12

Saran440:
Migrated from odoo v12 and fixed test script

PR #8, opened 2020-10-08
[13.0][FIX] rename hr_payroll module

davejames:
This PR renames the `hr_payroll` module to `payroll` to avoid conflicting with the Odoo Enterprise module of the same name.


davejames:
@Saran440 @pedrobaeza I have renamed the hr_payroll to payroll module as per the discussion in #5  

I also made a PR #7 to fix some problems with the hr_contract form, but have closed that PR and included those changes here instead. You can check that PR for some before/after screenshots.

Please can I trouble you for a review?

FernandoRomera:
Please, can your review this, I don't sure:

https://github.com/OCA/payroll/blob/456280e6504de1b152f5ea6e6fa5bc321a4a39d4/payroll/views/res_config_settings_views.xml#L61

name="module_account_accountant"
to
name="module_payroll_account"

https://github.com/OCA/payroll/blob/456280e6504de1b152f5ea6e6fa5bc321a4a39d4/payroll/views/res_config_settings_views.xml#L67

for="module_account_accountant"
to
for="module_payroll_account"

davejames:
You're right @FernandoRomera and there were also links to some localisations which have not yet been ported, so I have removed those for now.

AntoniRomera (approved):

FernandoRomera (approved):
Functional review, ok.

MRomeera (approved):
Ok

pedrobaeza:
You must replace at least in the POT file all the occurrences of `hr_payroll` by `payroll`, and later the bot will match them in each .po.

davejames:
It seems that .pot was included in .gitignore so the original pot file wasn't there to begin with. So I added that, and then replaced it with a freshly exported pot file. @pedrobaeza will the bot be OK with that?

I also removed .pot from .gitignore in line with other OCA repos

Saran440 (approved):
Thank you for your help :)
Functional Test :+1:

pedrobaeza (approved):
Thanks for the work

Issue #12, opened 2021-01-19
holiday.holiday_status_id.name => holiday.holiday_status_id.code - should be code, not name


vbezl:
https://github.com/OCA/payroll/blob/b88a195d92286f2ad3f7c88144fd048225df9175/payroll/models/hr_payslip.py#L353

Please change to:
"code": holiday.holiday_status_id.**code** or "GLOBAL",

nimarosa:
Refer to #14 pull request fixes the problem.

nimarosa:
#27 Fixed in 14.0 branch

nimarosa:
**We can close this issue the issue it's resolved in the 14.0 branch. Can be downgraded easily to other versions. **

PR #14, opened 2021-07-20
[13.0][FIX] Leave code and attendance calculation

nimarosa:
Hello, 

I noted that in hr_payslip.py the "code" it's refearing to the name and not the "code" in hr_holidays setted in the type field. I fixed that. 

Also in this commit, i changed the work_data to compute_leaves=False .- In some countries, the payslip should show the total amount of days (or hours) of the work schedule, and then, substract the absense days in another diferent salary_rule, so i think it's better to leave by default the calculation of the WORK100 woked day line to not consider leaves, so the user could change it in a setting (i'm developing this so i will do another pull request) or directly fetch the full work schedule days and hours and the leaves separately to work with the data in the salary rules. 

I think that is the best approach to support another countries and don't assume the user needs to discount the leaves in the worked_hours.

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -350,7 +350,7 @@ class HrPayslip(models.Model):
                     {
                         "name": holiday.holiday_status_id.name or _("Global Leaves"),
                         "sequence": 5,
-                        "code": holiday.holiday_status_id.name or "GLOBAL",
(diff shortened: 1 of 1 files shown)


nimarosa:
Sorry, when I put hr_holidays I mean hr_leave_type module

nimarosa:
Fixes #12

nimarosa:
Closed. Everything is included in the 14.0 branch PR.

PR #15, opened 2021-08-20
[IMP] payroll: Add login to maintainers key

AntoniRomera:


AntoniRomera:
@OCA/payroll-maintainers Anyone can approve me?

PR #17, opened 2021-09-03
[MAINTAINER] Proposal to be the maintainer of the addon

nimarosa:
Hello, i saw that you are looking for maintainers for the payroll addon. I'm developing the Argentinian Localization of the Payroll module, so i think i can maintain the full payroll addon and also add features to it. 

Here i submit the request with the change in the maintainers key in the manifest for your approval. 

Regards.

### DIFF
--- a/payroll/__manifest__.py
+++ b/payroll/__manifest__.py
@@ -9,6 +9,7 @@
     "summary": "Manage your employee payroll records",
     "license": "LGPL-3",
     "author": "Odoo SA, Odoo Community Association (OCA)",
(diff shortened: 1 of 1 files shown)


hitrosol:
I support this plan, we need to have payroll module in CE, especially in V14 and so on... Would you agree with this @pedrobaeza

pedrobaeza:
I'm afraid this is not the order: first there should be contributions, and then you can be the maintainer, so please do such contributions, and then, ask again.

PR #11, opened 2020-10-28
14.0 mig payroll

hilarak:
Migration to version 14.0 #6 Payroll

PR #18, opened 2021-10-13
[MIG] [14.0] Migration 'payroll' and 'payroll_account' 

nimarosa:
Here i PR my migration for version 14.0. Hope it can get merged so we can collaborate all together in the module. 

Regards.


pedrobaeza:
Please review instead #11. Anyways, you are not preserving commit history following https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0, and it should be only one module per PR.

PR #22, opened 2021-12-03
[13.0][OU-ADD] payroll,payroll_account: migration analysis

MiquelRForgeFlow:
As requested by @pedrobaeza 

It seems to me these module need migration scripts.


pedrobaeza (approved):
Thanks! cc @victoralmau

victoralmau:
Thank you! In my opinion it is correct although I think it is necessary to add the file `migrations/13.0.1.0.2/post-migration.py` to add `noupdate_changes.xml`, right?

pedrobaeza:
This is only the analysis. Now the migration scripts needs to be done.

PR #23, opened 2021-12-09
[13.0][OU-IMP] payroll: Add migration script since migration analysis.

victoralmau:
Add migration script since migration analysis (https://github.com/OCA/payroll/pull/22).

@Tecnativa TT32606

### DIFF
--- a/payroll/migrations/13.0.1.0.2/openupgrade_analysis_work.txt
+++ b/payroll/migrations/13.0.1.0.2/openupgrade_analysis_work.txt
@@ -1,0 +1,358 @@
+---Models in module 'payroll'---
+obsolete model report.hr_payroll.report_contributionregister (renamed to report.payroll.report_contributionregister) [abstract]
+obsolete model report.hr_payroll.report_payslipdetails (renamed to report.payroll.report_payslipdetails) [abstract]
+new model report.payroll.report_contributionregister (renamed from report.hr_payroll.report_contributionregister) [abstract]
+new model report.payroll.report_payslipdetails (renamed from report.hr_payroll.report_payslipdetails) [abstract]
+# NOTHING TO DO: abstract and transient views
+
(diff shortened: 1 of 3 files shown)


pedrobaeza (changes requested):
You have to make a copy of the `openupgrade_analysis.txt` file and call it `openupgrade_analysis_work.txt`, where you analyze by blocks the possible differences, or put `NOTHING TO DO` explaining the why's of each decision.

At least, I see 2 actions to do:

- There are `DEL` records with noupdate=1 that you should try to delete with the openupgradelib method (`delete_records_safely_by_xmlid` or similar)
- Some records that changes from noupdate=1 to noupdate=0 or vice versa, that you should also update its XML-ID with the corresponding method

but please check if anything more to do.

victoralmau:
Ok sorry, I have made the necessary changes after analyzing everything carefully.

payroll/migrations/13.0.1.0.2/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,358 @@
...
+
+NEW ir.module.category: payroll.module_category_payroll (noupdate)
+# NOTHING TO DO: Renamed
+
+NEW ir.rule: payroll.hr_contribution_register_rule_company (noupdate)
+NEW ir.rule: payroll.hr_payroll_structure_rule_company (noupdate)
+NEW ir.rule: payroll.hr_payslip_rule_company (noupdate)
+NEW ir.rule: payroll.hr_salary_category_rule_company (noupdate)
+NEW ir.rule: payroll.hr_salary_rule_company (noupdate)
+# DONE: post-migration: Add to noupdate_changes
```
pedrobaeza: If they are new records, why do you need to add them to the noupdate changes?

pedrobaeza (approved):

PR #27, opened 2022-01-25
[14.0] [FIX] Fixes in hr_payslip defaults and typo errors, more

nimarosa:
Hello, here i have some minor fixes that we are dragging from previus versions: 

- Fixed typo in leave calculation
- Worked days should not compute leaves by default, that is because leaves are computed in a separated work_hour entry, so we need to compute the full worked days and then in another salary rule we can subtract the leaves, otherwise it has no point in computing the leaves separated from the worked time. In many countries the default way is to show in the payslip the full worked hours (normal hours without leaves) and the subtract the corresponding leaves from the total. 
TODO: Also maybe we need to express the leaves as a negative number, i will update this PR with that. 
- I deleted the default starting data, because as i'm the developer of the only one Argentina Localization for payroll, i found out that this default values ends in the user mistaking in the salary rule creating. I think it's better to remove the default values and leave the user (or localization) create it.

I will add more minor fixes to this PR and i hope it can be merged this time... Payroll development for odoo has been abandoned for a long time and i hope we can continue developing this module. 

If you need maintainers i want to help. I'm actively developing payroll addons for my country and i have a lot of custom fixes and extra modules we could add to this repo. 

Please let me know if you need me to adjust anything in the PR for this to be merged.


pedrobaeza:
Put them as demo data?

nimarosa:
@pedrobaeza They are already in demo data, i removed them from production data because like i said, i dont see the point of populating a production database. Most of my users get confused and it really don't help because every country has they own rules for payroll. 

I can add them as demo data if you want, but there is a lot of demo data already loaded in demo.

pedrobaeza:
It's OK for me then. Please check pre-commit.

cc @AntoniRomera @apps2grow

nimarosa:
@pedrobaeza pre-commit ready.

nimarosa:
Okay so i added this to the demo data because the data is required for tests.

ghost:
Thank you @nimarosa for these improvements. I confirm that the demo data is installing correctly on demo, and not installed when not using demo. This change might require a migration script for existing instances, so it is good that the default branch is still 13.0.

Feel free to add yourself as a maintainer. Just add a new commit to this PR.

I am not really into testing, but in general I think it is good to write tests.

ghost (approved):

nimarosa:
@appstogrow mantainer key added

nimarosa:
@appstogrow Yes i don't write tests either. Maybe when we do more improvements we can write some. For now i think this is ready to merge if you want.

ghost:
@nimarosa In the OCA, one PR cannot have changes in more than one module. So in this PR you can only add yourself as a maintainer for the payroll module.

ghost:
Could you also try to squash the commits together into one commit?

nimarosa:
@appstogrow i've removed the mantainer key in payroll_account (will add later in another commit to this module). And also do the squash. Please tell me if it's okay because i don't often rebase branches so i might be wrong in the process. 

Hope everything it's okay to merge now.

AntoniRomera (approved):

ghost:
@nimarosa Yes I can see that the maintainer key is fixed. :+1: 
I see 10 commits now, I expected to see 1. [Here is an article about squashing.](https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec). Then `git push <your_repo> <your_branch_name> --force`

nimarosa:
@appstogrow I think it's ready now. I see only one commit.

ghost:
:+1:

nimarosa:
When can we merge this?

ghost:
@pedrobaeza Would you like to merge this PR?

PR #33, opened 2022-05-15
14.0-minor-changes-and-tweaks

nimarosa:
Hello, this PR adds more tweaks and fixes to the module. Like you know, i'm working intensively in this module for the Argentinian Localization for Payroll. 

Features and changes in this PR: 
- It adds a feature that takes into account the employee contract "start_date" for the calculations in the payslips. For example: If the employee started working before the "from_date" of payslip, payslip will only calculate work days and other data taking "start_date" as employee "from_date" for this payslip. This allows to don't have to make changes if the employee starts working in middle of a period/month. 
- Adds "struct_id" into onchange_employee method and support re-calculating all payslip worked_days and inputs when you change the "struct_id". This is useful if you have inputs and worked_days that calculates automatically depending of the rules included in the payslip. So we only calculate these lines when we need them. 
- Adds amount_qty to payslip inputs table. This give us an extra field which could be useful if we use inputs to fetch values which can have more than 1 unit. This do not modify any functionality and the amount_qty field added could be ignored completely if not used. 
- Add other states to payslip statusbar. 
- Reordered the hr_salary_rule view to have a more concise and practical view to create rules. 
- Minor view UX and UI improvements

If you like the PR please support it to get merged. Like you know, i will be making more PRs like this improving the module. 

Any suggestion is welcomed.

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -340,6 +340,10 @@ class HrPayslip(models.Model):
         ):
             day_from = datetime.combine(date_from, time.min)
             day_to = datetime.combine(date_to, time.max)
+            day_contract_start = datetime.combine(contract.date_start, time.min)
+            # only use payslip live from if it's greather than contract start date
+            if day_from < day_contract_start:
+                day_from = day_contract_start
 
             # compute leave days
             leaves = {}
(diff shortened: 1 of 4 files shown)


nimarosa:
@pedrobaeza If you can merge this it will be great. There are minor changes but adds a lot of value for some use cases in country localizations.

pedrobaeza:
Too much changes in only one commit. Please split them and put the proper commit message on each one. Also the view change doesn't seem very good if you new to add `<newline>` on it.

nimarosa:
@pedrobaeza <newline> is in the original code, i just moved it. 
I will close this PR in favour of a new one when i'm doing everythin in separated commits. Will PR soon.

PR #36, opened 2022-06-03
[IMP] payroll: add support for hr_public_holidays in payroll

nimarosa:
Hello, here a small PR. 

This add support to exclude the public holidays defined if you are using the OCA module hr_public_holidays. Before this change, payroll worked days calculations were no taking into account the public_holidays. 

Passing a context to the worked days method makes public holidays to be excluded in the calculations. If you are no using the module, it makes no difference or harm to pass this context. 

So this adds fully support for the module without making difference to the people who don't use it. 

@appstogrow @pedrobaeza What do u think?

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -339,18 +339,15 @@ class HrPayslip(models.Model):
         for contract in contracts.filtered(
             lambda contract: contract.resource_calendar_id
         ):
-
             # Adds support for the hr_public_holidays module.
             # This adds support to exclude public holidays from work days computation.
             # If you don't use the module, it don't make any difference or harm.
             contract = contract.with_context(
                 employee_id=self.employee_id.id, exclude_public_holidays=True
             )
-
             day_from = datetime.combine(date_from, time.min)
             day_to = datetime.combine(date_to, time.max)
             day_contract_start = datetime.combine(contract.date_start, time.min)
-
             # only use payslip day_from if it's greather than contract start date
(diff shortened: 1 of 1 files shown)


payroll/models/hr_payslip.py
```diff
@@ -339,9 +339,18 @@ def get_worked_day_lines(self, contracts, date_from, date_to):
...
+            # This adds support to exclude public holidays from work days computation.
+            # If you don't use the module, it don't make any difference or harm.
+            contract = contract.with_context(
+                employee_id=self.employee_id.id, exclude_public_holidays=True
+            )
+
             day_from = datetime.combine(date_from, time.min)
             day_to = datetime.combine(date_to, time.max)
             day_contract_start = datetime.combine(contract.date_start, time.min)
+
```
pedrobaeza: Don't add empty lines inside a method

pedrobaeza: And the same for the rest of empty lines that you introduce in this PR

nimarosa: The empty lines are there from previous versions of the module, I think the person who wrote it want it to separate different parts of the method. If you like i Can refactor the method and we fix that in this PR.

pedrobaeza: You are adding 3 empty lines in this PR . Please remove that ones.

nimarosa:
@pedrobaeza Ready 👍

pedrobaeza (approved):

PR #37, opened 2022-06-08
[14.0] [IMP] Remove hr_contract_advantage_template from payroll in favour of new module

nimarosa:
This PR completes the implementation of hr_contract_advantage_template currently without any use in the Payroll module. 

Before this PR, the user is able to set advantages_templates but there is no implementation of using this templates in the contract form like it should be. This PR adds the link between contract and advantages_templates and also add a new object that can be accessed form the salary rules in payslip: advantages.

With this object, the user can assign advantage_templates in the contract form and set the amount for this employee contract. Then there is a check for the amount field to be between upper and lower bound limits (defined in the template) and by default it uses the default_value field. 

This is very useful for the module because it allows the user to set specific parameters in each contract for every employee, so an employee can have his own values and parameters defined in his contract. It could be, for example, an extra fixed amount paid to a specific employee (and not to all employees), an attendance bonus, an extra amount for mobility. Also it can be used to set rates for specific salary rules like tribunal seizures applied for this employee and any other input that needs to be set in a employee level and not to the whole company. 

Hope we can get it merged. 

@appstogrow @pedrobaeza Like always, tell me what you think. 

I'm also preparing another PR implementing the contribution_register functionality, that like that one, it was incomplete in the module and it is not used. 

Regards.


nimarosa:
@appstogrow Can you review this so we can move further? Thanks in advance .

ghost:
Hi @nimarosa I have tested and it works.

Some suggestions:
1) Rename model to hr.contract.advantage (or hr.contract.payroll_advantage & hr.contract.payroll_advantage.template)
2) One python file for each model, not both in hr_contract_advantage_template.py
3) Add to the salary rule coding description (it took me some time to find out):
\# Note: Get worked_days / inputs / advantages by code, e.g. inputs.mycode
\# payslip: object containing the payslip [payslip, not payslips. I tested and got error message: Payslips object is not iterable]
4) External ID names:
access_hr_contract_payroll_advantage_hr_user
access_hr_contract_payroll_advantage_payroll_user
5) Squash the commits. The tag should be [IMP], not [14.0].

nimarosa:
@appstogrow Hey! Thanks for the comments, all of make sense to me. I did the changes and reworded the commit messages and squashed some trivial commits. 

Let me know if you think it's okay to merge or you want me to squash everything in one commit. 

Wait for your input. 

Edit: Finally i think it can be all in one commit. Correct me if wrong.

ghost:
@nimarosa One commit is fine. But the changes in hr_payslip.py are not in this commit.
What I tested yesterday, I have uploaded to https://github.com/appstogrow/payroll/commits/pr37
Could you please compare and see if anything more is missing?

nimarosa:
@appstogrow Yes you are right, something got missed in the last rebase. 

I added that changes and removed so not used code so travis test will pass now. 

Can you test and let me know? I made a mess with the last rebase haha.

ghost:
I made an advantage template with lower/upper limit = 0.
Trying to use the template in a contract gave an error: Advantage amount can't be greater than upper bound limit.
I think this error should not come when the upper limit = 0.

Otherwise the code works well.

nimarosa:
@appstogrow great observation. I did not tested that case. Will update next week and squash and we are ready.

mtelahun:
Hello @nimarosa
This is useful functionality but I think it should be in a module of its own. I am working on porting my own customization to allow this in a clean way.

nimarosa:
Hello @mtelahun, the advantage template model was part of payroll module all the time. This PR just finish implementing the functionality that was expected in the first time.

The author of the module created payroll advantages but not finished the implementation. 

Also this is just a first working implementation but I can be extended for more complex uses later. 

I don't agree this has to be a separate module. If the user don't want to use it, he just don't use it, because like I said the functionality was present but no usable. 

Also I think the payroll module must to add functionalities and don't remove the ones existing already. Making more separate modules don't make sense to me because payroll module must provide tools for doing payroll in any country, which are different from each other, so the most functions we can add the better. If a user don't need them, he can't just don't use them. 

Will be finishing this PR on Monday to allow a catch for the error described by @appstogrow

pedrobaeza:
Shouldn't this be a separate module?

nimarosa:
@pedrobaeza Hello, the model hr_contract_advantage_template is already part of the payroll module but it don't have any functionality (not implemented). 

This PR finish the implementation of the already present model. 

So I think It should be in the payroll module because it's the way it was planned to be. 

So we can merge this, or we should remove contract advantage templates from payroll because it don't make sense to have a table and views that don't do anything. 

I will prefer to merge this in payroll after I fix the bug in the exemptions but if you and @appstogrow  think it should be another module we can search the way because this implementation only works with the advantage templates present in payroll module.

pedrobaeza:
I prefer to extract the model `hr.contract.advantage.template` to a new module.

ghost:
I also prefer to extract the model `hr.contract.advantage.template` to a new module.

nimarosa:
@pedrobaeza @appstogrow Okay, can we make the module here in payroll repo? I think it has to be here and not in oca/hr because it's strictly related to payroll.

nimarosa:
Will update the PR removing the advantage template model and then create a new module.

pedrobaeza:
Yes, it can be here, but in OCA/hr-expense there's already a module for advantages that you may check.

nimarosa:
@pedrobaeza 

I can only find advance clearing modules in the repo (https://github.com/OCA/hr-expense/tree/14.0)

Contract advantage is intended to add specific employee contract values to be used in the salary rules, no to deal with salary advances clearing. 

Please tell me if you are referring to that module which is different or I'm looking to the wrong repo.

pedrobaeza:
Sorry, I have mixed both concepts. My fault.

nimarosa:
Closing, will make another PR removing advantage_templates and another adding the new module with this functionality.

mtelahun:
The reason I prefer that it be in a separate module is for the same 
reason you mentioned below: it has been sitting unfinished all this 
time, which means it's not really a necessary part of this module. Also, 
having it in a separate module allows multiple competing implementations 
to co-exist together.

I have started PR #43 to refactor the local dictionary passing to the 
salary rules. Once that is complete it will allow us to cleanly separate 
this kind of functionality without bloating the payroll module.

PR #40, opened 2022-06-17
[FIX] 14.0 - revert regression in pr 27

### LINKED ISSUE #38: Regression in behaviour of payroll module Worked Days Lines leave calculation
CC'ing developers involved in change
@nimarosa  @pedrobaeza  @apps2grow 

## Module
payroll

## Describe the bug
Hi guys, this change happened a few months ago but I only noticed it now when I updated my upstream sources and my unit tests started to fail.

PR #27 changed the sign of calculated hours and days of leaves to be negative. This is causing unit tests in my custom modules to fail.
I don't think the previous behavior was a bug, and even if it was I believe it's a bit late in the 14.0 life-cycle to be changing expected behavior now. Basically, this gratuitously changes previous behavior of the module without any tangible benefits to the wider community.

**Expected behavior**
That the module continues to function as it has before.

mtelahun:
Fixes #38


nimarosa:
As stated in the PR which made the change in the sign of leaves calculations: 

It makes no sense to sum the leaves in the worked hours table, when the entry WORK100 computes the full period of worked hours. 

So to make it clear: 
if you compute the total work hours in the WORK100 entry and then you sum the leaves, it leaves you with a invalid worked hours count. The correct approach to this is to substract the leaves  to the worked hours total, which is WORK100 entry. 

As the module works now, the WORK100 entry will calculate theoretical full worked hours and days, and then the entries for the leaves has to be a negative  number because the employee won't come to work that days and so we substract that hours and days to the total. If you make that entries a positive number, you are adding days that already has been computed. 

I don't understand your message in the issue, you said YOUR unit tests for custom modules fail after this changes. For me it looks that you have to change your unit tests and/or your custom modules. Maybe if you specify we can help you. 

Also you stated that there is no benefit for the user with this change but like I explained before, this has to be the behavior if we expect to calculate worked days and leaves as objects which sum must be the teal total worked days. 

Also, I think it's not late to make changes to the module. The payroll module was orphan a long time ago and it's development it's incomplete. I think version 14 it's a great version to start improving the module to allow backward compatibility and also a future migration to v15. 

Then you said something about the data which was moved to demo data. This is was it's not a good practice to populate a production database with this generic salary rules and structures. If your custom modules has some migration scripts that deletes that records or do something to the module, now you can pass of using this scripts because the task is added directly to payroll module. 

Also I think is not a very good practice to make your modules depend if scripts that changes things directly in the database, if I underhanded correctly. 

As stated I think this PR it's not required and not a fix. 

Please don't hesitate to answer if you have more questions about the new functionalities introduced. 

Regards.

mtelahun:
Hi Nicolas,

I don't have a problem with the changes, per se. My problem is that the 
behaviour of the module has changed in the middle of the 14.0 
life-cycle, without any clear benefit to the people who are already 
using it. My comment about my custom module failing was meant to 
highlight that this change has impacted me in production.  Yes, the 
payroll module has a lot of problems. However, anyone who has been using 
it in production has already worked around those problems and making the 
change now in 14.0 will break them (as it did for me).

My comment is about these two specific changes and not the overall 
changes that have gone in recently.

ghost:
Thank you @mtelahun for sharing your concerns.

I agree that the payroll module has a lot of problems. Therefore I suggest that we introduce a changelog like in [mis-builder](https://github.com/OCA/mis-builder/tree/15.0/mis_builder#changelog).

For data moving to demo, I think we should make a migration that removes the external id from ir.model.data. Then when people in the future migrates from 13.0 to 14.0 the data will not be deleted.

For the change in the sign of leaves calculations, I think the best way to announce it is in a changelog. Maybe it would need to be announced some time prior to the merging, so that people have the opportunity to fix code affected by the change.

We should also register that payroll & payroll_account is in alpha state, and comment in the repo that developments are going on so watch out. I would like to continue improving in 14.0, not wait until next version.

pedrobaeza:
For this specific case, and not knowing too much about the internals, what about adding a configuration parameter to have both behaviors?

nimarosa:
@mtelahun Okay i understand you. What I'm saying is that if people are using it and resolved some of the module drawbacks in his own custom modules, we have to encourage them to make PRs so we can make the module better. 

A lot of the changes I have been introducing I have them patched in my own payroll localization for my country and when we merge the changes in the main module I remove them in my custom one. 

Also I don't understand why you say that the changes don't provide clear benefit. I think that if people patched the behavior or use it the module knowing of the bug (maybe adapting the salary rules to it) people can update their modules according the fixes introduced here. We can't leave the module unfinished and not change it, that way the module will continue to be orphaned because people prefer to patch it for themselves and that do not encourage the people to use it and improve it, it's a toxic cicle haha. 

On the other hand, i agree with @appstogrow. I think it makes sense continue developing the 14.0 branch which for now it's the more used. Then a migration to 15 and further versions can be done quickly. I think we have to give support for at least 3 versions starting from 14.
I also agree about making a changelog and roadmap, in a way that don't slow down development of the module. 
I think we can make this module a very functional a useful one in no too much time, I have another PRs prepared. 

Finally, @pedrobaeza i will look into doing a configuration parameter to alternate the change. But i state that it really makes no much sense (in a payroll technical way) to sum the leaves instead of substrate them. 

Also I think this configuration parameters will not be much clear to the end user. The module, as it is, requires practice to learn how to use it (I have planned more documentation to help with that), so if we add this config it might be confusing for some users. In any case, if you and @appstogrow think it could be useful I can try to make it work in a future PR. 

But I clearly don't agree in merging this rollback changes because it affect the module development and improvement that has been made in the past months. I decided to contribute my changes to OCA because I believe in open source and this module it's a great one if we finish fixing bugs and implementing features. Merging this rollback will make difficult to continue the improvements I have been submitting because it changes the logic of the module, and forces to use a fork to support my country (and another's based in it) localization. 

Finally, I agree that we can make the demo data not removed after a migration. So it won't impact in version changes.

pedrobaeza:
I don't agree to revert as well. The PR was there for a while, and any cons or rejection had to be done then, not now. And also if the changes on top of this module are outside the OCA scope, then it's not very reasonable to ask for an inmovilism where other contributors are elevating the level of these modules a lot, and invest a lot of hours for the whole community.

@mtelahun if you want to preserve the old behavior, please contribute a PR adding such configurable parameter.

mtelahun:
I can live with a configurable parameter. I'll make another PR.

While I intend to be more involved in this repo in the future, please note that payroll is not like a reporting module or some "toy" feature. Regardless of how mature it is if it breaks suddenly "bad" things can happen.

pedrobaeza:
Well, all modules are affected by that things. Imagine a tax report bad calculated, or sales given with a bad price.

But if you are not involved actively in OCA in such critical things, then when you update your instances with latest code, you have to check the "diffs", especially with heavy custom layers on top.

mtelahun:
Yeah, you're right. If you want to play you have to pay :-)

Consider me chastened. I will engage more in HR, Payroll, and Helpdesk.

pedrobaeza:
Thanks for your comprehension. Eagering to see you around.

PR #41, opened 2022-06-19
[14.0][FIX] payroll unittest

mtelahun:
The unittest for the payroll module is broken in several ways:

1. It doesn't actually test anything in the payslip, just the states when various buttons are simulated
2. The payslip is created programatically, so no payslip values are loaded
3. The employee's contract is in "draft" state so even if it tried to load any values it wouldn't work because the contract isn't "open"
4. It modifies a salary input line, but doesn't verify it worked. It won't work because of the above mentioned reasons.

To ensure a minimal amount of testing:

- Make sure the employee's contract is in "open" state
- Use the "Form" testing object when creating a payslip to ensure the onchange_employee() method is run
- Actually test the payroll calculations are done correctly

### DIFF
--- a/payroll/tests/common.py
+++ b/payroll/tests/common.py
@@ -11,17 +11,146 @@ class TestPayslipBase(TransactionCase):
         super(TestPayslipBase, self).setUp()
 
         self.Payslip = self.env["hr.payslip"]
+        self.RuleInput = self.env["hr.rule.input"]
+        self.SalaryRule = self.env["hr.salary.rule"]
+        self.SalaryRuleCateg = self.env["hr.salary.rule.category"]
 
-        # Some salary rules references
-        self.hra_rule_id = self.ref("payroll.hr_salary_rule_houserentallowance1")
-        self.conv_rule_id = self.ref("payroll.hr_salary_rule_convanceallowance1")
-        self.prof_tax_rule_id = self.ref("payroll.hr_salary_rule_professionaltax1")
-        self.pf_rule_id = self.ref("payroll.hr_salary_rule_providentfund1")
-        self.mv_rule_id = self.ref("payroll.hr_salary_rule_meal_voucher")
-        self.comm_rule_id = self.ref("payroll.hr_salary_rule_sales_commission")
-        self.basic_rule_id = self.ref("payroll.hr_rule_basic")
-        self.gross_rule_id = self.ref("payroll.hr_rule_taxable")
-        self.net_rule_id = self.ref("payroll.hr_rule_net")
+        # Salary Rule Categories
+        #
+        self.categ_basic = self.SalaryRuleCateg.create(
+            {
+                "name": "Basic",
+                "code": "BASIC",
+            }
+        )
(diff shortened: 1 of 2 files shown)


rvalyi (approved):

pedrobaeza:
@nimarosa @apps2grow can you please check?

nimarosa:
@mtelahun Hello, it's a good PR. Payroll module has awful tests. 

I don't usually write tests so it's great to have a minimal functional one here. 

Your approach it's great and I agree with it. 

I suggest that you add unit tests that create the salary rules that will be used in the tests and don't use the demo ones because this tests will fail in production because a database without demo mode activated won't have these rules loaded. Also creating the salary rules in the tests give us a base to test this flow (creation of salary rules) in the future. 

If we add that and don't use demo está I think we can merge.

payroll/tests/common.py
```diff
@@ -10,13 +10,18 @@ class TestPayslipBase(TransactionCase):
...
+        self.Payslip = self.env["hr.payslip"]
+
         # Some salary rules references
         self.hra_rule_id = self.ref("payroll.hr_salary_rule_houserentallowance1")
         self.conv_rule_id = self.ref("payroll.hr_salary_rule_convanceallowance1")
         self.prof_tax_rule_id = self.ref("payroll.hr_salary_rule_professionaltax1")
         self.pf_rule_id = self.ref("payroll.hr_salary_rule_providentfund1")
         self.mv_rule_id = self.ref("payroll.hr_salary_rule_meal_voucher")
         self.comm_rule_id = self.ref("payroll.hr_salary_rule_sales_commission")
+        self.basic_rule_id = self.ref("payroll.hr_rule_basic")
```
nimarosa: Try to not use demo data for tests so it could be run in prod where no demo rules loaded.

mtelahun:
I agree it would be a good thing but the reality is that there are still 
so many tests, even in upstream Odoo, that require demo data that it is 
currently unrealistic to try to run unittests *without* demo data. In 
any event the goal of this PR was not to add new tests, but to fix the 
one that was already there.

But definitely a thing to keep in mind for future tests.

pedrobaeza:
> I agree it would be a good thing but the reality is that there are still
so many tests, even in upstream Odoo, that require demo data that it is
currently unrealistic to try to run unittests *without* demo data.

This is not true in the latest version. Odoo already removes the dependency on demo data in their tests. Anyways, that's independent from this, as you are referring 3 demo records in this PR diff. Just create them with their data in the test instead of adding new dependencies that will be needed to be removed in the future.

ghost (approved):

pedrobaeza (changes requested):
Pending of changing demo data to test data at least of the 3 new added rules.

mtelahun:
Will do

nimarosa (approved):

nimarosa (approved):

nimarosa:
Looks good to me now.

pedrobaeza (approved):
Thanks for switching everything

PR #42, opened 2022-06-19
[14.0][FIX] payroll - wrong quantity and rate during python computation of salary rule

mtelahun:
A side effect of the way the payslip line quantity and rate are set when
amount_select is set to "code" causes them to never evaluate to zero. When they
evaluate to 0 the side effect causes them to be set to 1.0 and 100.0, respectively.

    "result_qty" in localdict and localdict["result_qty"] or 1.0
    "result_rate" in localdict and localdict["result_rate"] or 100.0

### DIFF
--- a/payroll/models/hr_salary_rule.py
+++ b/payroll/models/hr_salary_rule.py
@@ -197,10 +197,16 @@ class HrSalaryRule(models.Model):
                 safe_eval(
                     self.amount_python_compute, localdict, mode="exec", nocopy=True
                 )
+                result_qty = 1.0
+                result_rate = 100.0
+                if "result_qty" in localdict:
+                    result_qty = localdict["result_qty"]
+                if "result_rate" in localdict:
+                    result_rate = localdict["result_rate"]
                 return (
                     float(localdict["result"]),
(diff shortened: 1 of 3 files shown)


ghost (approved):

nimarosa:
This change will cause an error when you don't set amount_qty and amount_rate because the system will expect to receive this values to convert them to float. 

I think that the previous approach is better because it looks for values and check existence before trying to use them. 

Also, which are the cases when you want this values evaluate to zero? I think default this to 1.0 and 100.0 it's the expected behavior if you don't set the values in the salary rule you expect them to be 1 and 100.

mtelahun:
No, actually it doesn't and I have added a second test to show that. 
Please re-read the PR. The problem was that if you set result_qty or 
result_rate to 0 in the salary rule the way the previous code evaluated 
those lines would reset them to their default values as if they weren't set.

nimarosa:
Yes you are right. I readed the PR in the phone yesterday and i didn't saw the conditionals you added, i thought you removed them but you changed that. 

Its okay for me. Approved.

nimarosa (approved):

PR #43, opened 2022-06-19
[14.0] [IMP] payroll: refactor of local dictionary handling in payslip for salary rule processing

mtelahun:
Any serious attempt at implementing a "real" payroll solution with Odoo will eventually require a way of injecting values into the payslip processing rules. At the same time we don't want people running full-blown python code against the Odoo ORM from a payslip rule. That is just an invitation for disaster during one of the most critical jobs of an ERP. So, the ideal solution will allow a module's author to inject discreet values (pieces of information) that can be easily calculated or accessed at run time but doesn't allow the salary rule any more access to the system than it already has.

This PR is an attempt to port my own customizations in this area in a more generalized form. The basic idea is to use the `baselocaldict` dictionary and the `BrowsableObject` python class to allow module authors to add their own nested
dictionaries that can be accessed during salary rule processing. There are two ways they can do this:
1. During loading of values for the payslip lines
2. Once for each contract processed in the payslip (for values that depend on the current contract being processed)

An example of the first kind can include but is not limited to:
1. The total number of contracts for the employee during this payslip
2. Calculation of values based on the previous periods payslip. For example, I use
this mecahnism to calculate the previous payslip's 'NET' value and generate a note
if there is a wide variance.

An example of the second kind is:
1. Hourly wage calculated based on the wage field in the contract

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -10,15 +10,31 @@ from odoo import _, api, fields, models, tools
 from odoo.exceptions import UserError, ValidationError
 
 
-# These classes are used in the _get_payslip_lines() method
-class BrowsableObject(object):
-    def __init__(self, employee_id, vals_dict, env):
-        self.employee_id = employee_id
+class BaseBrowsableObject(object):
+    def __init__(self, vals_dict):
+        self.__dict__["base_fields"] = ["base_fields", "dict"]
         self.dict = vals_dict
-        self.env = env
 
     def __getattr__(self, attr):
         return attr in self.dict and self.dict.__getitem__(attr) or 0.0
+
+    def __setattr__(self, attr, value):
+        _fields = self.__dict__["base_fields"]
+        if attr in _fields:
+            return super().__setattr__(attr, value)
+        self.__dict__["dict"][attr] = value
+
+    def __str__(self):
+        return str(self.__dict__)
+
+
+# These classes are used in the _get_payslip_lines() method
+class BrowsableObject(BaseBrowsableObject):
+    def __init__(self, employee_id, vals_dict, env):
+        super().__init__(vals_dict)
+        self.base_fields += ["employee_id", "env"]
+        self.employee_id = employee_id
+        self.env = env
 
 
 class InputLine(BrowsableObject):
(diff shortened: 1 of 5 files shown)


ghost:
@mtelahun This is interesting!
Do you have some example code to easily test "nested dictionaries that can be accessed during salary rule processing"?

mtelahun:
Yes. https://github.com/trevi-software/trevi-hr/blob/14.0/payroll_payslip_dictionary/models/hr_payslip.py

It's very useful for calculating all sorts of things and then making the values available to use in salary rules.

nimarosa:
Hi @mtelahun. It think it's a interesting solution. I think we can make in another module like I will do with contract advantages. That way we don't mess too much with the vanilla rules calculation process. 

You can inherit the function directly in another module and it will be much cleaner code to maintain and also much clearer for the end user who can choose if use it or not. 

Also If I understand correctly that this aporoach needs to be used by authors of other modules/localization and does not have a out of the box functionality, which I think make more sense in a separate module.

ghost:
@nimarosa Will this PR break the salary calculation for your clients?

_get_payslip_lines() is >100 lines. To change the logic in another module:
a) override the method by copying all >100 lines 
b) make a patch, or
c) rewrite the method into multiple smaller methods

Currently it is easier to do this implementation in payroll. To me it seems quite generic, which is what the payroll module is for.

nimarosa:
@appstogrow i didn't reviewed in extension because I think the author it's finishing committing changes. When it's ready I will review it, if it doesn't touch any of the values that payslip lines return I think we are safe.  

Edit: I've done a fast review and for now it looks we are safe. 

@mtelahun do you have pending commits for this to be ready? So we can review it. It looks like a nice refactoring too, congrats.

payroll/models/hr_payslip.py
```diff
@@ -440,35 +509,26 @@ def _sum_salary_rule_category(localdict, category, amount):
...
-            worked_days_dict[worked_days_line.code] = worked_days_line
-        for input_line in payslip.input_line_ids:
-            inputs_dict[input_line.code] = input_line
-
         categories = BrowsableObject(payslip.employee_id.id, {}, self.env)
-        inputs = InputLine(payslip.employee_id.id, inputs_dict, self.env)
-        worked_days = WorkedDays(payslip.employee_id.id, worked_days_dict, self.env)
-        payslips = Payslips(payslip.employee_id.id, payslip, self.env)
         rules = BrowsableObject(payslip.employee_id.id, rules_dict, self.env)
+        this_contract = BrowsableObject(payslip.employee_id.id, contract_dict, self.env)
```
nimarosa: I suggest changing this to "contract" or something more generic because "this_contract" will be easily confused.

payroll/models/hr_payslip.py
```diff
@@ -425,6 +433,67 @@ def get_inputs(self, contracts, date_from, date_to):
                 )
         return res
 
+    def get_payroll_dict(self, contracts):
```
nimarosa, line 449: It would be great to add a comment on how to inherit and add a dictionary in another module in this function. This will encourage developers to create new modules and integrations with another modules.

payroll/models/hr_payslip.py
```diff
@@ -425,6 +433,67 @@ def get_inputs(self, contracts, date_from, date_to):
...
+
+        baselocaldict = {
+            "payslip": payslips,
+            "worked_days": worked_days,
+            "inputs": inputs,
+            "payroll": payroll_dict,
+        }
+        return baselocaldict
+
+    def get_contract_dict(self, contract, contracts):
```
nimarosa, line 503: Also add examples and more documentation here.

nimarosa (changes requested):
Everything looks okay, i don't know if you have more changes to commit. If you do, please finish committing so we can review the final version. Also consider my comments to make this more documented and understandable. Aside of that, everything looks great.

mtelahun:
On 22/06/2022 16:53, Nicolas Rodriguez Sande wrote:
>
> I suggest changing this to "contract" or something more generic 
> because "this_contract" will be easily confused.
>
Yes, I thought about that, but "contract" is already used. I'm open to 
any other suggetions

nimarosa:
@mtelahun maybe current_contract?

mtelahun:
It's ready for review.

This should make it easier for other modules to inherit and modify the
payroll module. This solution allows other modules to arbitrarily
inject their own values into salary rules. Furthermore, payroll managers
will be able to continue using '.' (dot) notation to refer to these new
values. Module implementers can inherit and overload the following methods:

```
get_payroll_dict()  - for providing values under the generic `payroll.` object
get_baselocaldict() - for providing 'first-level' objects that are
                      available directly to salary rules. For the sake
                      of clarity the object string should start with
                      the module name. Example: if the module is named
                      payroll_foo the values should be under `payroll_foo.`
get_contract_dict() - for values that depend on the current contract being evaluated
                      These values will be available under `this_contract.`
```

In the inherited methods the modules should return a python dictionary of
discreet values and/or nested objects using a BaseBrowsableObject.
Example:
If the module name is my_module:
```
    res = super().get_baselocaldict(contract, contracts)
    res.update({
        "my_module": BaseBrowsableObject(
            "foo": 10.0,
            "bar": 25.0,
            "baz": BaseBrowsableObject(
                {
                    value1: 100.0,
                    value2: 150.0,
                }
            )
        )
    })
    return res
```

The above values can be accessed in salary rules as:
my_module.foo
my_module.bar
my_module.baz.value1
my_module.baz.value2

payroll/models/hr_payslip.py
mtelahun: current_contract sounds fine.

payroll/models/hr_payslip.py
mtelahun: In-line comments and actual examples in the code should help

payroll/models/hr_payslip.py
mtelahun: Comments in the code

payroll/models/hr_payslip.py
```diff
@@ -440,35 +545,26 @@ def _sum_salary_rule_category(localdict, category, amount):
...
-            worked_days_dict[worked_days_line.code] = worked_days_line
-        for input_line in payslip.input_line_ids:
-            inputs_dict[input_line.code] = input_line
-
         categories = BrowsableObject(payslip.employee_id.id, {}, self.env)
-        inputs = InputLine(payslip.employee_id.id, inputs_dict, self.env)
-        worked_days = WorkedDays(payslip.employee_id.id, worked_days_dict, self.env)
-        payslips = Payslips(payslip.employee_id.id, payslip, self.env)
         rules = BrowsableObject(payslip.employee_id.id, rules_dict, self.env)
+        this_contract = BrowsableObject(payslip.employee_id.id, contract_dict, self.env)
```
nimarosa: I suggest change this for "current_contract" to avoid confusion for the user.

payroll/models/hr_payslip.py
```diff
@@ -440,35 +545,26 @@ def _sum_salary_rule_category(localdict, category, amount):
...
+
+        baselocaldict = payslip.get_baselocaldict(contracts)
+        baselocaldict["categories"] = categories
+        baselocaldict["rules"] = rules
+        baselocaldict["this_contract"] = this_contract
```
nimarosa: Replace by "current_contract"

payroll/models/hr_payslip.py
```diff
@@ -483,6 +579,10 @@ def _sum_salary_rule_category(localdict, category, amount):
 
         for contract in contracts:
             employee = contract.employee_id
+            contract_dict = payslip.get_contract_dict(contract, contracts)
+            baselocaldict["this_contract"] = BrowsableObject(
```
nimarosa: Replace by "current_contract"

payroll/models/hr_payslip.py
```diff
@@ -534,6 +634,7 @@ def _sum_salary_rule_category(localdict, category, amount):
                 else:
                     # blacklist this rule and its children
                     blacklist += [id for id, seq in rule._recursive_search_of_rules()]
+            baselocaldict["this_contract"] = {}
```
nimarosa: Replace by "current_contract"

nimarosa (changes requested):
For me it looks great. 
Pending replacing "this_contract" with "current_contract" or something similar to make more comprehensible by the user. 

Also i suggest editing the salary_rule comment help information (the text that appears when you create a new salary rule and you select python) to reflect the available objects with this change. If you guys think that this text it's too long or poor explanatory i suggest we add in the salary_rule view a section when we explain all available objects that can be used in salary rules.

nimarosa (approved):
For me it's ready to merge. @pedrobaeza @appstogrow Could you review this and merge it if you think it's okay?

PR #45, opened 2022-06-21
[14.0] [IMP] payroll: Remove contract_advantage_template model and new module

nimarosa:
Like we discussed in closed PR #37, here i remove the hr_contract_advantage_template model from the payroll module and its views. 

Will make another PR with the new module implementing this functionality so the user can choose to use it or not. 

Please @pedrobaeza can we merge that as discussed? 

Regards.


pedrobaeza:
I think it's better to propose both in the same PR, as they are closely tied.

nimarosa:
@pedrobaeza Okay will make the module and add it here. One thing, how can i bootstrap all the readme and  dynamic oca data that the module requires? Or it's okay to just add the module without it? 

For example, readme.rst dynamic generation.

pedrobaeza:
You can have the readme folder with the fragments, and when merged, the README.rst will be generated, or you can pre generate it in advance installing locally OCA/maintainer-tools.

ghost:
> I think it's better to propose both in the same PR, as they are closely tied.

Is it ok to change multiple modules in the same **COMMIT**? Or would this make it impossible to migrate one module at a time to a new version?

pedrobaeza:
It's not a problem, as the git command for preserving the history extracts the folder.

nimarosa:
Hello, I'm preparing the module right now but I face a issue: 

The previous PR made changes in payslip to make contract advantages available in salary rules localdict. 

I think the problem will be that even if I inherit the function and then do a super() , the way that the original code constructs localdict will make me override almost everything to get the data in salary rules. This will make problems if people use this new module and then payroll module changes (like we are changing it recently) so it will need constant maintenance. 

If we ship this functionality inside payroll module it makes fewer changes to the payslip code so it would be more easy to maintain and don't affect the users that don't use this functionality. 

Would you reconsider merging this to payroll instead? @appstogrow @pedrobaeza because i think it does not any harm to people who don't use contract advantages and will make it easy for us to finish the implementation here than in another module. 

Wait for your comments I almost have the module ready but in order to continue I need to know what you think.

mtelahun:
Hello @nimarosa

This is what PR #43 is for. This use-case is common when localizing Odoo 
payroll. Rather than have all these use-cases in the  payroll module it 
makes more sense to abstract away the interfaces and have the details 
implemented in separate modules. For example, for this use-case you 
would just have to overload get_contract_dict():

```

from odoo.addons.payroll.models.hr_payslip import BrowsableObject

def get_contract_dict(self, contract, contracts):

     res = super().get_contract_dict(contract, contracts)

     ...

     adv = {

         "code1": foo,

         "code2": bar,

     }

     res.update({

         "advantages": BrowsableObject(adv),

     })

     return res

```

In the salary rule you would use:

` result = this_contract.advantages.code1`

nimarosa:
@mtelahun i really like that refactoring because it's true. In my own localization I overrided some methods and this way It's a lot cleaner. 

I think I will wait for your PR to be finished to continue with this one. 

Also before migrating your PR I will need us to wait for me to update my localization to this new way, so my clients don't loose support after that change. 

Hope to see your PR finished. If you need help let me know.

nimarosa:
Changing to draft until #43 is merged.

nimarosa:
Added module. Will test it when PR #43 is merged.

mtelahun:
I think so too. There are a few other places where I think hr.payslip 
can be re-factored some more, but they can wait. As far as the changes 
going in, you shouldn't see anything different. Only the internal 
behaviour has changed. Existing modules that depend on payroll shouldn't 
break.

ghost:
Update payroll_contract_advantages/security/ir.model.access.csv
- hr.group_hr_user
- payroll.group_payroll_user

ghost:
@nimarosa Are you on vacation now or would you have time to fix the security groups mentioned above?

nimarosa:
@appstogrow Hello, I will. I did not had time recently because I'm very busy with a project. 
Next week I will finish this PR and continue with the other one that is open.

nimarosa:
@appstogrow The security groups are fixed if you want to test. I can't figure out why travis fails. If you could help will be great.

ghost:
@nimarosa 
pr45 does not have the latest code of branch 14.0.
get_contract_dict() is missing in payroll/models/hr_payslip.py.
Try git rebase 14.0 (?)

nimarosa:
@appstogrow you are right.. I created the branch before that was merged. Will rebase today.

nimarosa:
@appstogrow Just rebased, please tell me if you think is okay, i have a mess with old branches haha

ghost (changes requested):
In payroll_contract_advantages/models/hr_payslip.py get_contract_dict():
- self.ensure_one()
- res.update({"advantages": BrowsableObject(self.employee_id, advantages_dict, self.env)})

In payroll/models/hr_payslip.py get_contract_dict():
- Is it correct that the code should do nothing?

nimarosa:
@appstogrow 

1. Thanks for the suggestion, will update soon. 

2. That change was introduced by @mtelahun and the use of that function is to be inherited or overrided by other modules like this one. But out of the box it does nothing as you said. I think it's good to have it because I allows fast implementation of new payslip salary rule objects.

mtelahun:
> In payroll_contract_advantages/models/hr_payslip.py get_contract_dict():
> 
>     * self.ensure_one()
> 
>     * res.update({"advantages": BrowsableObject(self.employee_id, advantages_dict, self.env)})
> 
> 
> In payroll/models/hr_payslip.py get_contract_dict():
> 
>     * Is it correct that the code should do nothing?

Yes, it's there to be inherited and overridden by other modules.

nimarosa:
@appstogrow Fixed your comments, thanks for that. In my development environment is working, if you want to test it and find anything please let me know. Finally, think we are ready to merge.

ghost:
I suggest you rebase with 14.0 and squash the commits. Then I will test again.

nimarosa:
@appstogrow Rebase done.

nimarosa:
@appstogrow i think you can only merge patches? @pedrobaeza could you help us here? 

Edit: I think maybe is because you are not maintainer in the new module? Want me to try? I don't think it will let me merge anything but can try.

pedrobaeza:
/ocabot merge major

I think you three should apply as PSC of payroll

PR #50, opened 2022-08-11
[IMP] payroll: allow the use of positive values in leave days/hours c…

mtelahun:
This is controlled by a new system parameter in Payroll Settings. The default is off. But when set to true leave days and hours will no longer appear as negative numbers. There are two reasons why someone might want this.

First, the idea that the sum of worked days and leave days should equal the number of actual days worked is **NOT** correct. A trivial example is if we are tracking overtime.  For example, let's say Sally worked only two days this month. The first day she worked a normal 8 hours, but the second day she worked an extra 2 hours of overtime (total 10 hours). In this case the worked days lines would look like this:

| Description                |   Code        | No. of Days | No. of Hours |
| --------------------------- |  -------------- | ---------------- | ----------------- |
| Normal Hours 100% |  WORK100 |           2        |           16       |
| Overtime                   |   OT            |           1        |             2       |
|Sum                           |                    |           3        |                     |

So according to this simplistic view Sally worked for 3 days this month when this is patently false. This is only one trivial example and there are many more. For example, some leaves may be paid leaves that by law cannot be deducted from pay (so for payroll purposes it should be treated as a worked day).

The second reason is more of a philosophical objection but leads back to the first. The payroll code should only give back or display the magnitude (or absolute value) of a number. It's up to the human (through salary rules) to decide whether it should be treated as a negative or positive number. The programmer should not force his/her own mental model onto the user. This is **WRONG.** It introduces unnecessary hidden complexity. With the current behavior a salary rule that calculates worked days will look like this: 

`result = worked_days.WORK100.number_of_days + worked_days.SICK.number_of_days`

The following more closely resembles the actual operation the user is trying to do:

`result = worked_days.WORK100.number_of_days - worked_days.SICK.number_of_days`

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -388,6 +388,9 @@ class HrPayslip(models.Model):
         @return: returns a list containing the leave inputs for the period
         of the payslip. One record per leave type.
         """
+        leaves_positive = (
+            self.env["ir.config_parameter"].sudo().get_param("payroll.leaves_positive")
+        )
         leaves = {}
         calendar = contract.resource_calendar_id
         tz = timezone(calendar.tz)
@@ -407,14 +410,20 @@ class HrPayslip(models.Model):
                     "contract_id": contract.id,
                 },
             )
-            current_leave_struct["number_of_hours"] -= hours
+            if leaves_positive:
+                current_leave_struct["number_of_hours"] += hours
+            else:
+                current_leave_struct["number_of_hours"] -= hours
             work_hours = calendar.get_work_hours_count(
                 tz.localize(datetime.combine(day, time.min)),
                 tz.localize(datetime.combine(day, time.max)),
                 compute_leaves=False,
             )
             if work_hours:
-                current_leave_struct["number_of_days"] -= hours / work_hours
+                if leaves_positive:
+                    current_leave_struct["number_of_days"] += hours / work_hours
+                else:
+                    current_leave_struct["number_of_days"] -= hours / work_hours
         return leaves.values()
 
     def _compute_worked_days(self, contract, day_from, day_to):
--- a/payroll/models/res_config_settings.py
+++ b/payroll/models/res_config_settings.py
@@ -7,3 +7,9 @@ class ResConfigSettings(models.TransientModel):
     _inherit = "res.config.settings"
 
     module_payroll_account = fields.Boolean(string="Payroll Accounting")
+    leaves_positive = fields.Boolean(
+        config_parameter="payroll.leaves_positive",
+        string="Leaves with positive values",
+        help="Values for leaves (days and hours fields) "
+        "should be positive instead of negative.",
+    )
--- a/payroll/tests/__init__.py
+++ b/payroll/tests/__init__.py
@@ -1,5 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from . import test_browsable_object
+from . import test_hr_payslip_worked_days
 from . import test_hr_salary_rule
 from . import test_payslip_flow
--- a/payroll/tests/common.py
+++ b/payroll/tests/common.py
@@ -14,6 +14,7 @@ class TestPayslipBase(TransactionCase):
         self.RuleInput = self.env["hr.rule.input"]
         self.SalaryRule = self.env["hr.salary.rule"]
         self.SalaryRuleCateg = self.env["hr.salary.rule.category"]
+        self.Contract = self.env["hr.contract"]
 
         # Salary Rule Categories
         #
--- a/payroll/tests/test_hr_payslip_worked_days.py
+++ b/payroll/tests/test_hr_payslip_worked_days.py
@@ -1,0 +1,106 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from datetime import date, datetime
+
+from odoo.tests.common import Form
+
+from .common import TestPayslipBase
+
+
+class TestWorkedDays(TestPayslipBase):
+    def setUp(self):
+        super().setUp()
+
+        self.LeaveRequest = self.env["hr.leave"]
+        self.LeaveType = self.env["hr.leave.type"]
+
+        # create holiday type
+        self.holiday_type = self.LeaveType.create(
+            {
+                "name": "TestLeaveType",
+                "code": "TESTLV",
+                "allocation_type": "no",
+                "leave_validation_type": "no_validation",
+            }
+        )
+
+    def _common_contract_leave_setup(self):
+
+        # I put all eligible contracts (including Richard's) in an "open" state
+        self.apply_contract_cron()
+
+        # Create the leave
+        self.LeaveRequest.create(
+            {
+                "name": "Hol11",
+                "employee_id": self.richard_emp.id,
+                "holiday_status_id": self.holiday_type.id,
+                "date_from": datetime.combine(date.today(), datetime.min.time()),
+                "date_to": datetime.combine(date.today(), datetime.max.time()),
+                "number_of_days": 1,
+            }
+        )
+
+    def test_worked_days_negative(self):
+
+        self._common_contract_leave_setup()
+
+        # Set system parameter
+        self.env["ir.config_parameter"].sudo().set_param(
+            "payroll.leaves_positive", False
+        )
+
+        # I create an employee Payslip
+        frm = Form(self.Payslip)
+        frm.employee_id = self.richard_emp
+        richard_payslip = frm.save()
+
+        worked_days_codes = richard_payslip.worked_days_line_ids.mapped("code")
+        self.assertIn(
+            "TESTLV", worked_days_codes, "The leave is in the 'Worked Days' list"
+        )
+        wdl_ids = richard_payslip.worked_days_line_ids.filtered(
+            lambda x: x.code == "TESTLV"
+        )
+        self.assertEqual(len(wdl_ids), 1, "There is only one line matching the leave")
+        self.assertEqual(
+            wdl_ids[0].number_of_days,
+            -1.0,
+            "The days worked value is a NEGATIVE number",
+        )
+        self.assertEqual(
+            wdl_ids[0].number_of_hours,
+            -8.0,
+            "The hours worked value is a NEGATIVE number",
+        )
+
+    def test_leaves_positive(self):
+
+        self._common_contract_leave_setup()
+
+        # Set system parameter
+        self.env["ir.config_parameter"].sudo().set_param(
+            "payroll.leaves_positive", True
+        )
+
+        # I create an employee Payslip
+        frm = Form(self.Payslip)
+        frm.employee_id = self.richard_emp
+        richard_payslip = frm.save()
+
+        worked_days_codes = richard_payslip.worked_days_line_ids.mapped("code")
+        self.assertIn(
+            "TESTLV", worked_days_codes, "The leave is in the 'Worked Days' list"
+        )
+        wdl_ids = richard_payslip.worked_days_line_ids.filtered(
+            lambda x: x.code == "TESTLV"
+        )
+        self.assertEqual(len(wdl_ids), 1, "There is only one line matching the leave")
+        self.assertEqual(
+            wdl_ids[0].number_of_days, 1.0, "The days worked value is a POSITIVE number"
+        )
+        self.assertEqual(
+            wdl_ids[0].number_of_hours,
+            8.0,
+            "The hours worked value is a POSITIVE number",
+        )
--- a/payroll/views/res_config_settings_views.xml
+++ b/payroll/views/res_config_settings_views.xml
@@ -41,6 +41,20 @@
                             </div>
                         </div>
                     </div>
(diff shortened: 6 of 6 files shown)


nimarosa:
@mtelahun

 Okay let's discuss this approach. Anyways if you want to make the config parameter it's okay but please change the help text that states that worked_day_lines should never be negative, it's confusing for the user and maybe not in your country, but in the vast majority of latin countries and some Europe ones I worked with your approach is not true. 

Here we are not trying to force anything because like you said it depends of the coder of the salary rule which approach to use. 

Regarding your example I have some thoughts: 

1. If you are tracking overtime the numbers must be positive ones, but I don't understand how it affects you because the negative numbers are for leaves. Are you tracking overtime in the leave module? If you are, so then is you who are forcing your mentality here because the leaves (or holidays) module is for leaves and holidays and not for tracking overtime. Instead there is no public module for tracking overtime out of the box, we have a private one for that. So in that point of view I don't know why overtime has anything to do with leave calculation to you. 

2. In your example table you pointed that if the employee has 2 hours of overtime the table will show 1 day. Why is that? If the work schedule is 8 hours then the day calculation will show float 0.25 not 1 day as you said. I don't understand your logic there, you are welcome to explain it better if you want. Again I don't get why negative numbers have anything related to this example. 

3. Regarding your second reason, again I don't know in your country but in the European and Latin countries payroll is used to only make sum operations and not subtraction ones. That's why there is a accounting principle that states that the sum of all entries should always be equal to the balance of the equation. 

Following your example why you need a salary rule that calculates worked days if the table have already given you the values. 

If you need full worked days you just simply do: 

result_qty = worked_days.WORK100.number_of_days 

There you already have the full working schedule without need of summing the sick days to get the actual number. So with the current behavior you or the payroll officer doing the formulas have the values he needs without having to do any operation. 

Then if you need to deduct the sick days in the payslip, you Just do: 

result_qty = worked_days.SICKDAYS.number_of_days 

and the value will come negative (because of the current behavior) so you can make a rule that deducts sick days from the full worked salary without having to do any operation. 

---- 

The logic of this behavior is to make the module more user oriented, and not dev oriented. Payroll officers will want a variable that returns the value they are looking for without having to make calculations even if there are simplistic like this example. 

The way of thinking is that you always want to have the full worked days number and show it in the payslip and then deduct the leaves, holidays, sum the overtime, etc. If you have all as positive numbers it makes confusing for the user and also don't make logic sense. 

In the most payroll systems and businesses I saw (and believe me I saw a lot of them) this is the default behavior. You sum the days that sum to the Normal working schedule (WORK100) and deduct the ones that has to be deducted (sick days, etc.). 

Also regarding your statement that there are days that are leaves but are paid ones, you can deduct it and pay it again with a salary rule. That's the way it is in normal payroll systems. You inform all the normal summing information and then you deduct what it has to be deducted. If some deduction is a payed one, you use another salary rule to sum it again. This is because the payslip MUST show the employee all the calculations to make a transparent payslip, and i a lot of countries this is a MUST / OBLIGATION by law. If you inform the days all in positive number and deduct all from WORK100 (like it was before) your are not being clear with the employee about how much he schedule must have been and how much has been deducted 

A user doing payroll will expect that behavior in a payroll system. Again, we are not forcing any logic here, only adapting the module to what a user expects. 

If we as a community want people migrating to odoo and also make odoo a industry grade system, we have to make our modules work like the system users expect, not because we want, because it's better for adoption. That's why this change was introduced in the first place, because is the normal logic of payroll systems. 

I understand that your use case don't adapt to this, and that's why you can always make a module that override the functions like you want. 

But stating that a way of thinking is only trying to force a mentality it's not okay. I have several years working in a lot of payroll systems and that's why I stated contributing to this module, to make it competitive with the other ones. And that is a function that the user will expect. 

I read again your comment and I can't figure out why this functionality affects you. I don't see correlation in your example and the problem your are proposing. 
As you said, you can adapt your salary rules to contemplate this or even override it in your custom module. 

Finally I understand there are a lot of different use cases for every odoo module but we can't expect to adapt to all of them. We have to make the modules the way that the most number of users would expect and that's because other system work like that it's not a coincidence. 

We can make the config parameter id you want but I recommend you removing the blinded statements in the help texts that affirms that "it should never be negative" because is confusing and not necessary for it to be there. Don't wring a approach discussion to the code. 

Regards.

ghost:
Thank you @mtelahun for this pr with good description and tests.

I tested this pr in a fresh database with demo data. In Payroll Settings, when I tick off "Worked days non-negative" and click save, the setting is not saved.

nimarosa:
@appstogrow @pedrobaeza i invite you to join this topic and give your opinion, then we can vote. 

I don't have problem with a config parameter but maybe we can discuss each other our opinions of this module development approach.

nimarosa:
Also if we are going to move forward with this I think the default behavior must be the actual one (negative numbers) so the user can choose to enable this if he want but we should not force it by default because like I said the expected behavior is the actual one.

ghost:
From a human perspective,
- A leave does not have a negative duration.
- A leave is reducing days worked.
So registering leaves under "Worked Days" in the payroll, it is not clear to a user whether a leave should have positive or negative value.

I thought solving this with a configuration parameter was a good idea. But if the user can change the parameter, the salary rules will have to read the configuration parameter to compute correctly. Maybe this is not a good idea after all.

What about renaming "Worked Days" to "Worked Days and Leaves", and add a selection field to show either "Work" or "Leave" on each line?
- That would be clear to the user.
- Calculations could use the selection field to determine whether to use plus or minus.

nimarosa:
@appstogrow 

I'm not agree with you guys in that one. From a human perspective a LEAVE is a day (or hours) an employee didn't work. So a leave is a deduction (subtracting) from the full work days schedule. Note that the WORK100 entry calculates the FULL worked days and hours of the employee without taking into account the leaves, so then, in the table and also in the payslip, you subtract the leaves. 
I don't think it's no clear because in any payroll system i've seen the leaves ALWAYS are negative, because there are leaves = days not worked. I don't think the user should make a decision of if the leave should be positive or negative because is always negative.   

For a more detailed example we can think in a employee that has a 8 hour work schedule from monday to friday. This will be in a 30 days month a working schedule of 20 days and 160 hours. So that is the full schedule the employee must have worked if him has no leaves. 
Then suppose the employee was absent because any reason, you have to deduct that days from the full working schedule so your table will be like this: 

| Description | Worked Days | Worked Hours | 
| ------------- | ------------- | ------------- | 
| WORK100  | 20  | 160 | 
| Unjustified leaves | -1  | -8 | 
| Sum real worked days| 19 | 152 | 

So in the payslip you will have the following salary rules if the monthly salary is 10.000: 
- Theoretical worked days (gross salary) => $ 10.000
- Unjustified leaves => $ -500 (gross / theoretical worked days * leave) 
- Total => $ 9.500

All other worked_days lines will be positive (worked days, overtimes, etc) , the only ones that will be negative are the leaves. So the user don't have to decide any of that from a user point of view its clear. 

-----

On the other hand, we can make the title more descriptive if you want. We can use "Worked Days and Leaves" if you guys think is more clear that way. I think a selection field is redundant because the leaves come with a code and a description, the description is fetched from the leave configuration in the holidays module, so a line that is a leave will have a description like: "Unjustified leaves" "Holiday leave" "Sick days leave", so it's already clear for the user what is each line representing.
But if you think is necessary we can make the selection implementation. For me is redundant because the description is already clear but it should no be much work doing it.

Actually the module only fetches worked days and leaves, so there is only two types of records that could be there. I have another modules that add another type of worked_days inputs like overtime, holidays, public holidays. But that's not covered out of the box by this module (it's in my plans bringing that integrations to here in the future but for now, the system only fetches that two records and all of them brings the description of what it is).

I don't think there is such generic salary_rule that needs to decide anything. In most of the payroll implementation that i've seen and implemented, you will have a unique salary rule for each type of leave , or a general one , according to your needs in the payslip report. 

-----

I think that the main confusion @mtelahun is having with this topic is that he is considering that WORK100 entry already reduced the days worked, when in reality it is showing the full work schedule for this period and employee and then it's job of the payslip computation to deduct that amount with a salary rule. Like i said earlier, if you deduct directly the leaves from the WORK100 entry, 

1. You will need to make calculations in all salary rules to determine the full work schedule. That is not expected by the used in almost any payroll system because a lot of calculations in payroll are based on the full work schedule, not the deducted schedule.
2. It makes a not transparent payslip because you don't print to the employee the real worked time so he can see the deductions made, you deducted from start so i don't quite see why is the benefit of it. 

-----

I think we can make the parameter but as @appstogrow said, it will be confusing to the user and also salary_rules will need to adapt to that setting. If we don't have problem with that, we can do it. 

I have more than 10 businesses consuming this module as a dependency that would not expect that change, but i can warn them. All of that 10 businesses (being some latin american and european) expected the module to work like is working. So i think it's something i'm not getting of why @mtelahun views this as some weird function.

mtelahun:
Like I said in the PR, there is a reason why someone would prefer 
absolute values and not negative and positive numbers. I will also point 
out that until you changed the behavior it has always been positive, 
starting with the initial import of the module from Odoo SA. As I said 
in our previous discussion I don't have a problem with it being negative 
by default. And frankly, I don't want to waste any more cycles on this 
issue. I just wanted to set the record straight for **why** someone 
might only want absolute (ie. non-negative) values.

mtelahun:
Sorry about that. It's fixed now.

mtelahun:
As I said previously it's fine with me if the default is negative as 
long as I have a way to turn it off.

Having said that, let's not muddy the issue here. The default behavior 
has always been positive, until you changed it.

mtelahun:
Yes, in the long run it would be better to record worked days and leaves 
separately. But let's leave that for the future. Right now i just need a 
way to bring back the previous behavior.

ghost (approved):
Now I could save the setting "Worked days non-negative". Then leaves are computed with positive value in the payslip.

Consider this text in the Payroll Settings:
Worked Days
- Leaves with positive values
- - Values for leaves (days and hours fields) should be positive instead of negative.

mtelahun:
I have already stated the justification for this PR and I will leave it 
at that. But you raise some points that are worth discussing outside the 
scope of this PR. What follows is a somewhat philosophical rant on 
software development.

The arguments you make are an excellent illustration of my second point. 
We as programmers should not force our own mental model on the user. We 
should strive to make it as easy as possible for the user to make the 
software fit into **their** mental model and not try to make them fit 
their mental model to our software. User here includes the person 
writing the salary rules. As far as I'm concerned if the implementer 
needs a programmer to write or modify salary rules then we have failed.

You are correct in describing the situation as **you** see it. For your 
use-case it makes sense. However, in other instances it does not:

* the payroll is based on actual hours worked (ie. based on attendance). 
In this case a leave should not be deducted from worked days because 
there is no attendance record anyways

* the payroll is based on the work schedule but some leaves are 
protected by law and should therefore be treated as a worked day. In my 
country pregnant women are allowed to take 1 day medical leave that 
cannot be deducted from their salary.

* another example that is applicable to most countries is paid annual 
leave. If the payroll is based on work schedule then you can't subtract 
it from the worked days or else you have to find some other way to add 
back the deducted amount, which goes back to my point about forcing the 
user to work **our** way.


I hope this is a clearer explanation for why I think negative numbers 
for leaves are wrong.

mtelahun:
Thanks, that's a much better choice of words.

nimarosa:
@mtelahun as I said earlier, I don't try to force anything, I only wanted to (as you also have done) point the why the change that was implemented so we keep a record of this discussions in case this topics emerge again. 

Like I said before at the beginning, I have no problem with making a configuration parameter so the user could choose the way it works. 

As you said in my country and some others pregnant woman also have paid leaves. In that case the law and the vast majority of payslips deduct the amount of leaves and then (if they are paid ones) paid it again in another line with the description "paid Pregnancy leave" so it makes clear in the payslip that the employee was absent but then we reimburse the amount because is a legal paid leave. 

I agree with you I don't want to discuss this topic much more and I agree with the parameter. 

I suggest the field and help text to be something like this: 

Use absolute leaves and calculation in payslips
- If you enable this field, all leaves (days and hours fields) will be treated as absolute numbers and should be positive instead of negative. 

So we can make clear to the user about the functionality of the setting. It's a bit longer but more explained.

nimarosa (approved):

ghost:
I see `worked_days_positive` eleven times in the code. @mtelahun would you like to change it to `leaves_positive`?

Also consider the description in res_config_settings_views.xml
- Display non-negative values for leaves in payslip worked days lines. (suggested by you)
- Values for leaves (days and hours fields) should be positive instead of negative. (suggested by me)
- If you enable this field, all leaves (days and hours fields) will be treated as absolute numbers and should be positive instead of negative. (suggested by @nimarosa)

mtelahun:
Ok, leaves_positives it is then. As for the wording I already took your 
suggestion and changed it Saturday.

Thanks,

Mike.

nimarosa:
@mtelahun Hello, now the payslip refactoring PR is merged, we need to adjust the merging conflicts in this PR. Probably just rebase with 14.0 and make the same changes in the new codebase and we can merge this.

mtelahun:
Rebase on top of 14.0

PR #53, opened 2022-08-16
[14.0][IMP] payroll: send email with payslip, fix payslip form view, better payslip line views

ghost:
This PR has 3 commits:
[IMP] payroll: send email with payslip
[FIX] payroll: payslip form view: name attributes for inheritance
[IMP] payroll: payslip line: better views, new field payslip_run_id

### DIFF
--- a/payroll/views/hr_payslip_line_views.xml
+++ b/payroll/views/hr_payslip_line_views.xml
@@ -20,6 +20,7 @@
                 <field name="total" />
                 <field name="amount_select" invisible="1" />
                 <field name="register_id" invisible="1" />
+                <field name="payslip_run_id" optional="hide" />
             </tree>
         </field>
     </record>
@@ -53,6 +54,18 @@
             </form>
         </field>
     </record>
+    <record id="hr_payslip_line_view_pivot" model="ir.ui.view">
+        <field name="name">hr.payslip.line.pivot</field>
+        <field name="model">hr.payslip.line</field>
+        <field name="mode">primary</field>
+        <field name="priority" eval="20" />
+        <field name="arch" type="xml">
+            <pivot>
+                <field name="salary_rule_id" type="row" />
+                <field name="total" type="measure" />
+            </pivot>
+        </field>
+    </record>
     <record id="hr_payslip_line_view_search" model="ir.ui.view">
         <field name="name">hr.payslip.line.select</field>
         <field name="model">hr.payslip.line</field>
@@ -65,6 +78,7 @@
                 />
                 <field name="amount_select" />
                 <field name="slip_id" />
+                <field name="payslip_run_id" />
                 <group col="8" colspan="4" expand="0" string="Group By">
                     <filter
                         string="Salary Rule Category"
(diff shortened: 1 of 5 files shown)


nimarosa:
@appstogrow 

This looks great to me, if you run pre-commit before pushing it will pass the checks. Congrats on the email functionality, i was thinking in doing something like this time ago. 

I think runboat fails because you are missing the action (action_report_payslip) declaration.

ghost:
Any idea why the positive/negative leave test doesn't create a Worked Days line with the leave?

nimarosa:
> Any idea why the positive/negative leave test doesn't create a Worked Days line with the leave?

Maybe @mtelahun can answer this since she submitted the setting PR?

mtelahun:
Arghh! Default resource.calendar defines a 5-day work week so the test fails if you run it on Saturday or Sunday.

nimarosa:
@mtelahun maybe you can make a PR and make the test run with fixed dates? Or maybe you can create a full week calendar? 

Because that test will affect all PR tests if we make PRs in the weekend.

nimarosa:
@appstogrow tests should be fixed with last merge. I think you gonna have to rebase.

ghost:
It should be ready now.

payroll/views/hr_payslip_line_views.xml
```diff
@@ -20,6 +20,7 @@
                 <field name="total" />
                 <field name="amount_select" invisible="1" />
                 <field name="register_id" invisible="1" />
+                <field name="payslip_run_id" />
```
nimarosa: I think we should keep this field invisible or better with the parameter optional="hide", so the user can show it or hide it depending of the situation. Payslip lines are often viewed from payslip form.

payroll/views/hr_payslip_line_views.xml
```diff
@@ -105,11 +119,8 @@
     >
         <field name="name">Payslip Computation Details</field>
         <field name="res_model">hr.payslip.line</field>
-        <field name="view_mode">tree,form</field>
-        <field name="domain">[]</field>
-        <field name="context">
-            {'default_slip_id': active_id,'search_default_slip_id': active_id}
-        </field>
+        <field name="view_mode">tree,pivot,form</field>
+        <field name="domain">[('slip_id', 'in', active_ids)]</field>
```
nimarosa, line 123: Not sure what this change do?

nimarosa: the domain one

payroll/views/hr_payslip_line_views.xml
```diff
@@ -65,6 +78,7 @@
                 />
                 <field name="amount_select" />
                 <field name="slip_id" />
+                <field name="payslip_run_id" />
```
nimarosa, line 81: Is there a case when the payslip_line_ids form is not displayed from the smart button in the payslip? There is no menu to access the full payslip_lines views. If i'm not wrong, then having a search for payslip_run_id will always return the same records in that tree view.

nimarosa (changes requested):

payroll/views/hr_payslip_line_views.xml
ghost: Select multiple payslips and click Action - Payslip Computation Details to see the lines of all the selected payslips (not just the first payslip). active_ids works in domain, but unfortunately not in context.

UPDATE: I don't need to change this action bound to Payslip. I created a new action bound to Payslip Batch.

payroll/views/hr_payslip_line_views.xml
ghost: The user may use the Action menu to navigate to payslip lines list view. Then the user can search.

I just want to see pivot totals for a batch. So I made an action menu for Payslip Batch.

nimarosa: Great I didn't know that this action existed haha. Great

nimarosa (approved):

nimarosa:
Okay so we are ready to merge this. 

/ocabot merge minor

PR #55, opened 2022-08-17
[14.0][FIX] payroll_account: log a warning if no move lines are generated for the payslip

mtelahun:
The payroll_account module tries to confirm the account moves for the payslip without verifying that there are actually any account move lines. This causes the account module to throw an exception. I ran into this when one of my module's tests (that has nothing to do with payroll accounting) kept triggering this failure. The payroll_account module test seems to imply that this failure is expected but I think this is wrong because A) it causes other modules that do not know about payroll_account to fail and B) this is a downstream exception in another module and the payroll_account module should not be relying on an exception being thrown in another module.

### DIFF
--- a/payroll_account/models/hr_payroll_account.py
+++ b/payroll_account/models/hr_payroll_account.py
@@ -1,7 +1,11 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
+import logging
+
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
+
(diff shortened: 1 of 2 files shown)


nimarosa:
Yes I agree with you and replicated the error in a clean installation. 
Will test it and if everything turns okay i agree to merge.

nimarosa (approved):

nimarosa:
Tested. Let's try. 

/ocabot merge patch

PR #56, opened 2022-08-19
[FIX] payroll: Fix Refactored _compute_name to contemplate more than …

nimarosa:
This fix is for the new _compute_name method introduced in last PR. The method was contemplating only one record, so when its called from payslip_run we get an error. 

This fixes it. Will merge directly since it's a easy fix and can be a problem for production installations.

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -794,16 +794,17 @@ class HrPayslip(models.Model):
         self.company_id = self.employee_id.company_id
 
     def _compute_name(self):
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as direct patch:

/ocabot merge patch

nimarosa:
@pedrobaeza oops, we merge at the same time, hope the bot understands haha

pedrobaeza:
Hehe, I aborted mine.

PR #57, opened 2022-08-21
[FIX] payroll: use a custom working calendar so tests don't fail on weekends

mtelahun:


nimarosa:
Okay let's merge this. 

/ocabot merge patch

Issue #29, opened 2022-02-10
Salary computation not working as expected


olivierb2:
## Module
payroll

## Describe the bug
Adding a manual value into "Salary Computation" doesn't recalculate other lines.

## To Reproduce
**Affected versions**: tested on Odoo 13.0 and 14.0

1. Create new payslip
2. Click on "Compute Sheet" button to populate Salary Computation
3. Click on "Edit"
4. Click "Add a line"
5. Add new item "Bonus"

**Expected behavior**
While "Bonus" has code than BASIC, I expected to have all other lines computed, like GROSS which is supposed to be categories.BASIC + categories.ALW.

nimarosa:
Yes, it's a known bug. I have a patch for it, will make a new PR on that on the 14.0 branch to fix it.

housemateguy:
Will there be a fix for this in 13.0 ?

olivierb2:
Same for me, patch for 13.0 is also useful for me as my company is currently still on 13.0.

nimarosa:
@appstogrow Do you want to work in this with me? As you are also doing a localization for payroll. This is a known error but i think it's difficult to patch and maybe we will have to change some of the core functionality.

ghost:
@nimarosa Does your patch work for 14.0 also after the changes done in the latest PRs? If you share your patch, anyone can contribute. This issue is low priority for me, but probably I can review and give feedback.

@olivierb2 and @housemateguy, when will you migrate to 14.0?

olivierb2:
I'm not sure to migrate to 14.0 at this stage, I have reconcile feature missing on my testing database, I'm not sure why Odoo remove features each time they do new version :-(

ghost:
@olivierb2 Did you try the new OCA module [account_reconciliation_widget](https://github.com/OCA/account-reconcile/tree/14.0/account_reconciliation_widget)?

olivierb2:
Not sure it's the best place to discuss about this, but yes, I tried it. But logic is completely changed and looks much less convenient than before.

nimarosa:
> @nimarosa Does your patch work for 14.0 also after the changes done in the latest PRs? If you share your patch, anyone can contribute. This issue is low priority for me, but probably I can review and give feedback.
> 
> @olivierb2 and @housemateguy, when will you migrate to 14.0?

I have a partial patch. Will make a PR so we can work in it. But it has to be tested because it changes the way of making the calculations. 

[...]

PR #60, opened 2022-08-24
[IMP] payroll: Improve views, add child functionalities, allow compute of name

nimarosa:
Hello friends, 

This PR brings the following changes to the payroll module: 

1.  Improvements to the views and forms of hr_payslip and hr_payslip_line 

2. As is done in enterprise, payslip_lines should not be available to edit manually, also the table should not allow introducing the lines manually. This is fixed in that PR. As the table can no be edited anymore, this allows us to do any computation and conditions to the table to allow functionalities like the next one. So #29 is now fixed since the module now don't let manually add lines, they should be always computed by the module. 

3. Support child and parent salary rules in form and allowing more functionality. Now child salary rules are differentiated in the computation table (with muted color and bold the parent ones, see images), also if you click on one line to open it's associated form, you can now see a table with the associated lines of that payslip (this is done doing a recursive relation by the field parent_rule_id). 
Before this change, the functionality of child and parent rules was trivial, since they has no impact in the calculation and lines generated are not differentiated in any way. This allows the user to use child and parent salary rules as auxiliar calculations for the parent salary rule and also see them in the table or in the associated payslip_line form allowing an analytic a easy way to review a payslip. Finally we have a flag boolean field on top of the table that allow the user to hide/show the child rules as his wish. 

4. Now details_by_salary_rule_category is filtered by appears_on_payslip. That allow to use that table to have a view of the actual lines that compose the payslip and it's report. TODO: Probably we should change the name of the tab, i accept suggestions about this. The method i leave the same name (details_by_salary_rule_category) to preserve backward compatibility with private modules that might be using it. 

5. I introduce the variable "result_name" available for computation in python computed salary rules. The result_name variable allows the user to use a python code to compute a name, and this name is propagated to the payslip_line name. This is useful for use cases when a payslip_line should change its name depending of a condition, or when you want to display some employee specific information in the line name. If variable is not used or it's value is "False", the default salary_rule name will be used as always. 

Hope you like the new changes, and i accept suggestions if you have an idea to make it even better. I think that they are very useful functionalities for the module to have. Hope you test it soon and let me know.

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -234,6 +234,11 @@ class HrPayslip(models.Model):
         compute="_compute_details_by_salary_rule_category",
         string="Details by Salary Rule Category",
     )
+    dynamic_filtered_payslip_lines = fields.One2many(
+        "hr.payslip.line",
+        compute="_compute_dynamic_filtered_payslip_lines",
+        string="Dynamic Filtered Payslip Lines",
+    )
     credit_note = fields.Boolean(
         string="Credit Note",
         readonly=True,
@@ -251,13 +256,24 @@ class HrPayslip(models.Model):
     payslip_count = fields.Integer(
         compute="_compute_payslip_count", string="Payslip Computation Details"
     )
+    hide_child_lines = fields.Boolean(string="Hide Child Lines", default=False)
+
+    @api.depends("line_ids", "hide_child_lines")
+    def _compute_dynamic_filtered_payslip_lines(self):
+        for payslip in self:
+            if payslip.hide_child_lines:
+                payslip.dynamic_filtered_payslip_lines = payslip.mapped(
+                    "line_ids"
+                ).filtered(lambda line: not line.parent_rule_id)
+            else:
+                payslip.dynamic_filtered_payslip_lines = payslip.line_ids
 
     @api.depends("line_ids")
     def _compute_details_by_salary_rule_category(self):
         for payslip in self:
             payslip.details_by_salary_rule_category = payslip.mapped(
                 "line_ids"
-            ).filtered(lambda line: line.category_id)
+            ).filtered(lambda line: line.category_id and line.appears_on_payslip)
 
     def _compute_payslip_count(self):
         for payslip in self:
@@ -632,10 +648,11 @@ class HrPayslip(models.Model):
                 localdict["result"] = None
                 localdict["result_qty"] = 1.0
                 localdict["result_rate"] = 100
+                localdict["result_name"] = None
                 # check if the rule can be applied
                 if rule._satisfy_condition(localdict) and rule.id not in blacklist:
                     # compute the amount of the rule
-                    amount, qty, rate = rule._compute_rule(localdict)
+                    amount, qty, rate, computed_name = rule._compute_rule(localdict)
                     # check if there is already a rule computed with that code
                     previous_amount = (
                         rule.code in localdict and localdict[rule.code] or 0.0
@@ -653,11 +670,12 @@ class HrPayslip(models.Model):
                     result_dict[key] = {
                         "salary_rule_id": rule.id,
                         "contract_id": contract.id,
-                        "name": rule.name,
+                        "name": computed_name and str(computed_name) or rule.name,
                         "code": rule.code,
                         "category_id": rule.category_id.id,
                         "sequence": rule.sequence,
                         "appears_on_payslip": rule.appears_on_payslip,
+                        "parent_rule_id": rule.parent_rule_id.id,
                         "condition_select": rule.condition_select,
                         "condition_python": rule.condition_python,
                         "condition_range": rule.condition_range,
--- a/payroll/models/hr_payslip_line.py
+++ b/payroll/models/hr_payslip_line.py
@@ -16,6 +16,15 @@ class HrPayslipLine(models.Model):
     payslip_run_id = fields.Many2one(
         "hr.payslip.run", related="slip_id.payslip_run_id", string="Payslip Batch"
     )
+    child_ids = fields.One2many(
+        "hr.payslip.line", "parent_line_id", string="Child Payslip Lines"
+    )
+    parent_line_id = fields.Many2one(
+        "hr.payslip.line",
+        string="Parent Payslip Line",
+        compute="_compute_parent_line_id",
+        store=True,
+    )
     salary_rule_id = fields.Many2one("hr.salary.rule", string="Rule", required=True)
     employee_id = fields.Many2one("hr.employee", string="Employee", required=True)
     contract_id = fields.Many2one(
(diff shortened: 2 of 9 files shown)


nimarosa:
I can't figure out why tests fails. The expected values to unpack now are 4 not 3. So the error in Travis is the expected functionality. Someone with more experience of tests can take a look?

mtelahun:
Hi @nimarosa 
Can you break this up into one change per PR please. It's hard to review as one big PR. Thanks.

nimarosa:
@mtelahun actually the changes are correlated, so I can't break it in multiple PR because one change implies the other. 

The only one change I can break is the name one, but this change didn't make several modifications. 

I suggest you try the PR in a dev environment and you will see the changes, then you can see the code. That's why I made a list of the changes.

nimarosa:
@appstogrow do you want to review this? It contain the comments that I made in the issue thread.

mtelahun:
@nimarosa There is a good reason why the OCA guidelines say one change per PR. It reduces the burden on the reviewers and there is less chance that bugs slip by. It's fine if one patch depends on another one being merged first. Also, I feel uncomfortable that such an extensive change goes in without any additional tests.

mtelahun:
Also, it would be more helpful if the PR description said **why** the changes are being made and not just a description of the changes.

nimarosa:
@mtelahun the majority of the changes are on the view level. Then I only added a new function for the new table in the form. But in reality there is not a huge change in terms of the logic or anything like that.

I know what OCA guidelines say, but as I said, it makes no sense to make more than one PR if one change will not work with the other one. 

I will fix the tests to test the new name functionality. This is the only new functionality, the rest are views changes and a new function. 

Let's wait to see what others have to say. Like I said I can make the name functionality in another PR since that is the only one change that didn't depend on others. All the child functionality is tied up to the views changes. 

If you want suggest a way to break this in more PRs and I will do it if it makes sense for us.

nimarosa:
> Also, it would be more helpful if the PR description said **why** the changes are being made and not just a description of the changes.

@mtelahun You can find the discussion of that changes in the details_by_salary_rule_category issue. 

I think the description is clear, I didn't want to make a long desc that confuse. Also the description says why the changes are being made. Anyways here is why: 

- Child and parent support. Today the child and parent rules have no functionality since the module takes all rules and apply them by sequence. With this change, we diff the lines generated by that rules with colors in the table, allow to hide or show them on demand and also show them in the payslip_line contextual form, so it can be used as auxiliar calculations and we make a visual representation of them to the user. 

- we added a appears-on-Payslip filter to the method details by rule category, this is what was intended by the module if you look at the views, prior to version 11 odoo allowed to group and filter on view level, after odoo 11 this function was removed and that's why now it shows the same information that line_ids. So we added a filter at domain level to recover the function to show only lines that appear on payslip there. 

- Then these two changes are tied up to all the views changes that you see in the PR and also to make that change possible we made line_ids not editable like is done in enterprise payroll module, this is why it makes no sense to edit or add payslip lines manually, it should always be calculated.

Hope I clarified for you. If you want more information you can look at the issue.

payroll/models/hr_payslip.py
```diff
@@ -653,11 +670,12 @@ def _sum_salary_rule_category(localdict, category, amount):
                     result_dict[key] = {
                         "salary_rule_id": rule.id,
                         "contract_id": contract.id,
-                        "name": rule.name,
+                        "name": computed_name != "False" and computed_name or rule.name,
```
mtelahun: Should this be False or "False"?

nimarosa: The method do a str(result_name) and that returns a "False" string, that's why the condition takes both cases if it is False or "False".

mtelahun:
Can we return an empty string in that case? What if the developer 
**wants** to return "False" as a string? This is the kind of thing that 
causes confusion (and bugs) down the line.

nimarosa:
@mtelahun If we return an empty string we can't do the condition to use computed name or the rule one. 

To use this you will use in the rule Python code as follows: 

result_name = employee.job_id.name or False 

- This will change the line name to the job name (as an example) or return False if not job_id is set. That's why then we check if the variable is False or "False" because the user will probably want to be false if no name set. If we don't make that comparison then if the employee has not job_id we get an error. That's why we make the "False" approach.

Anyways What we can do is not make the str() in the rule and do it in the payslip. So if the variable is Boolean false, we do nothing and use the rule name, and if it is any other thing, we do the str().

Will make a new approach and update the PR.

mtelahun:
I'm getting the following stack trace when I try to run it. One of the rules has child rules. When I remove it from the salary structure I don't get a stack trace:
```
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 386, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/auto/addons/payroll/models/hr_payslip.py", line 367, in compute_sheet
    for line in self._get_payslip_lines(contract_ids, payslip.id)
  File "/opt/odoo/auto/addons/payroll/models/hr_payslip.py", line 655, in _get_payslip_lines
    amount, qty, rate, computed_name = rule._compute_rule(localdict)
```

nimarosa:
@mtelahun could you post the full trace for me to replicate it? Thanks

nimarosa:
@mtelahun Tests are passing now. The problem was that i only considered the result_name variable for python computed rules, so if you have another type of rules it brings an unpack error because the value is not passed. Now for "percentage" and "fixed" type of rules, the value of result_name is by default False, meaning that result_name only can be computed in python salary rules. 

This should fix tests, and your use case. Also taken your suggestion and made the default value of result_name boolean False. So first we use this value to know what name to use, and then if the value is not False, we do the str(result_name). Working now.

nimarosa:
Added tests for the new child_lines functionality. So we can have test coverage and also be sure it's working like it does in future changes.

nimarosa:
@appstogrow @mtelahun This is ready for review. Please let me know if i can merge it so i can open others PR with an updated branch

ibuioli (approved):

ibuioli (approved):

ghost:
"result_name" works.
"Details By Salary Rule Category" would currently more correctly be named "Salary rules appearing on payslip". But we can keep the old tab name for now.
I haven't tested child rules. It would be nice with a tab description explaining the purpose.

ghost (approved):

nimarosa:
@appstogrow i can make the tab name change in this PR. 

What do you mean with "a tab description"?

nimarosa:
@appstogrow I changed the name of the tab and also added the info description in salary_rules form to explain the use. 
Will rebase, wait for tests and merge. 

After the merge, i will start with the refactoring we were talking about in #51 so you can use it in your PR.

nimarosa:
PR approved, tested and working. 
Merging mejor version since the changes are important.

Keep posted for more new changes coming soon. 

/ocabot merge major

PR #62, opened 2022-09-03
[IMP] Payroll: Payslip refactoring and misc functionalies

nimarosa:
Hello, this is the first refactoring PR of a set of refactorings i will be submitting to the module. 

I will do it in separate PRs to make them more easy to test, review and merge, and also to separate concerns. 

This PR, introduces the following changes: 

1. Refactoring of functions _get_payslip_lines and _get_localdict that are the main functions involved in payslip calculation. With this changes, the behavior is the same, but separating of concerns and refactoring in more short functions, give us the ability to inherit the functions in any external module in a easy and straightforward way. We now have _compute_payslip_line new function, that is used for calculating ONE line, so if we need to make changes inside the lines loop we can do it in this function.

2. Change in functions returns: Now you will note that the functions in general return localdict every time. This is useful because if we can edit localdict we can create any number of functions inheriting the main one and just change the localdict to get the data in easily. The _get_payslip_lines function now returns localdict, lines_dict meaning lines_dict the original "result_dict". In the prior implementation, since the function only returns result_dict, there was no easy way to inherit the function and change something that is saved inside localdict. Like i said, the behaviour of the calculation is the same but we now have more useful return values. 

3. Changes in _get_baselocaldict function, now this function creates the baselocaldict entirely. Previously part of the localdict was created in this function and part was hardcoded in the _get_payslip_lines function. Now we have this function complete and here we set the defaults and different objects for the localdict. 

4. I moved BaseBrowsableObject and its childs to a new file, so we reduce the length of the payslip file and also have a separate file to edit this local classes. 

5. I removed advantage_templates unused functions that were in hr_contract, since this functionality was moved to a new module. #45 

6. Now the function that sum the categories, is in the hr_salary_rule_category file, this means that we can inherit it and change it if we need in an external module. Also this is an example of the implementation of returning localdict, note the function takes the localdict as param, make the sum, and return localdict, so parent function can continue the loop. 

7. Introducing new field: "compute_date", which is a date field that will be automatically set when the payslip is computed, this way we can keep track of the last computed date. This is useful for cases when the user don't know if the computation saved is the last one and also for keeping a track of every computation. 

8. States: Now we are using the "verify" state, that in this module implementation was unused. I checked how the enterprise version deals with states and they use the "verify" state to be automatically set when the payslip is computed. This way, when you compute the payslip the state is changed to "verify" which indicates that the payslip has been computed and is waiting for the user to check it and confirm it. When the state is "verify" you can still compute the payslip again, the only difference is that when the state is not "draft" all fields in the payslip become not editable ones: So this is useful when you have different roles doing payroll. An HR user can compute the payslips and then he can not modify them (since the become "verify" state) and has to wait for the HR Manager to approve them. Anyway if the user want to change any date in the payslip he can just press "return to draft" and then compute again. 
I think this makes a lot of sense. Maybe we have to work in permissions groups to allow/disallow return to draft function for different roles. But now, we are using all the states defined in payslip, since before we only use draft, and confirm ones.

9. Finally we have a lot of tweaks in views, removal of unused or deprecated view functionalities and the addition of a lot of filters to some models. Payroll module has outdated views so i think we can even improve it more like it was done in previous PRs. But here we only added/improve what is related to this PR. Feel free to suggest changes because views are very important in this module and improve them is a priority.

*Important note:* 
While refactoring the files i noted that the feature introduced in #43 was a very useful one for external modules. More than I noticed when it was merged. I think we have to make more documentation about the new objects present in localdict "payroll" and "current_contract" since they allow to inject values directly to the localdict without having to inherit any other function. 

Now this PR allows more easy inheriting of module functions but i think for external modules like the one in #51 by @appstogrow this objects are the best approach to inject new values to localdict. We should use them every time we can, because that way we don't interact with payroll module functions and makes a more clear approach. Henrik, i think you can study this functionality but i think you can use this approach in your PR to inject the localdict values you need in your new module. Please check it. Anyways now you can inherit the main functions without problem but like i said, i think this is more useful approach to use. 

For now, this is the first PR of a series of refactoring ones. Please check and i wait for your comments/suggestions. 
Regards.

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -523,8 +523,8 @@ class HrPayslip(models.Model):
             "total": rule_total,
         }
         # sum the amount for its salary category
-        localdict = rule.category_id._sum_salary_rule_category(
-            localdict, rule_total - previous_amount
+        localdict = self._sum_salary_rule_category(
+            localdict, rule.category_id, rule_total - previous_amount
         )
         # create/overwrite the rule in the temporary results
         key = rule.code + "-" + str(localdict["contract"].id)
@@ -651,6 +651,17 @@ class HrPayslip(models.Model):
             }
         )
         return res
+
+    def _sum_salary_rule_category(self, localdict, category, amount):
+        self.ensure_one()
+        if category.parent_id:
+            localdict = self._sum_salary_rule_category(
+                localdict, category.parent_id, amount
+            )
+        localdict["categories"].dict[category.code] = (
+            localdict["categories"].dict.get(category.code, 0) + amount
+        )
+        return localdict
 
     def _get_employee_contracts(self):
         return self.env["hr.contract"].browse(
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -9,110 +9,19 @@ from pytz import timezone
 from odoo import _, api, fields, models, tools
 from odoo.exceptions import UserError, ValidationError
 
-
-class BaseBrowsableObject(object):
-    def __init__(self, vals_dict):
-        self.__dict__["base_fields"] = ["base_fields", "dict"]
-        self.dict = vals_dict
-
-    def __getattr__(self, attr):
-        return attr in self.dict and self.dict.__getitem__(attr) or 0.0
-
-    def __setattr__(self, attr, value):
-        _fields = self.__dict__["base_fields"]
-        if attr in _fields:
-            return super().__setattr__(attr, value)
-        self.__dict__["dict"][attr] = value
-
-    def __str__(self):
-        return str(self.__dict__)
-
-
-# These classes are used in the _get_payslip_lines() method
-class BrowsableObject(BaseBrowsableObject):
-    def __init__(self, employee_id, vals_dict, env):
-        super().__init__(vals_dict)
-        self.base_fields += ["employee_id", "env"]
-        self.employee_id = employee_id
-        self.env = env
-
-
-class InputLine(BrowsableObject):
-    """a class that will be used into the python code, mainly for
-    usability purposes"""
-
-    def sum(self, code, from_date, to_date=None):
-        if to_date is None:
-            to_date = fields.Date.today()
-        self.env.cr.execute(
-            """
-            SELECT sum(amount) as sum
-            FROM hr_payslip as hp, hr_payslip_input as pi
-            WHERE hp.employee_id = %s AND hp.state = 'done'
-            AND hp.date_from >= %s AND hp.date_to <= %s
-            AND hp.id = pi.payslip_id AND pi.code = %s""",
-            (self.employee_id, from_date, to_date, code),
-        )
-        return self.env.cr.fetchone()[0] or 0.0
-
-
-class WorkedDays(BrowsableObject):
-    """a class that will be used into the python code, mainly for
-    usability purposes"""
-
-    def _sum(self, code, from_date, to_date=None):
-        if to_date is None:
-            to_date = fields.Date.today()
-        self.env.cr.execute(
-            """
-            SELECT sum(number_of_days) as number_of_days,
-             sum(number_of_hours) as number_of_hours
-            FROM hr_payslip as hp, hr_payslip_worked_days as pi
-            WHERE hp.employee_id = %s AND hp.state = 'done'
-            AND hp.date_from >= %s AND hp.date_to <= %s
-            AND hp.id = pi.payslip_id AND pi.code = %s""",
-            (self.employee_id, from_date, to_date, code),
-        )
-        return self.env.cr.fetchone()
-
-    def sum(self, code, from_date, to_date=None):
-        res = self._sum(code, from_date, to_date)
-        return res and res[0] or 0.0
-
-    def sum_hours(self, code, from_date, to_date=None):
-        res = self._sum(code, from_date, to_date)
-        return res and res[1] or 0.0
-
-
-class Payslips(BrowsableObject):
-    """a class that will be used into the python code, mainly for
-    usability purposes"""
-
-    def sum(self, code, from_date, to_date=None):
-        if to_date is None:
-            to_date = fields.Date.today()
-        self.env.cr.execute(
-            """SELECT sum(case when hp.credit_note = False then
-            (pl.total) else (-pl.total) end)
-                    FROM hr_payslip as hp, hr_payslip_line as pl
-                    WHERE hp.employee_id = %s AND hp.state = 'done'
-                    AND hp.date_from >= %s AND hp.date_to <= %s AND
-                     hp.id = pl.slip_id AND pl.code = %s""",
-            (self.employee_id, from_date, to_date, code),
-        )
-        res = self.env.cr.fetchone()
-        return res and res[0] or 0.0
-
-    def rule_parameter(self, code):
-        return self.env["hr.rule.parameter"]._get_parameter_from_code(
-            code, self.dict.date_to
-        )
+from odoo.addons.payroll.models.base_browsable import (
+    BaseBrowsableObject,
+    BrowsableObject,
+    InputLine,
+    Payslips,
+    WorkedDays,
+)
 
 
 class HrPayslip(models.Model):
     _name = "hr.payslip"
     _inherit = ["mail.thread", "mail.activity.mixin"]
-    _description = "Pay Slip"
+    _description = "Payslip"
     _order = "id desc"
 
     struct_id = fields.Many2one(
@@ -160,8 +69,6 @@ class HrPayslip(models.Model):
         states={"draft": [("readonly", False)]},
         tracking=True,
     )
-    # this is chaos: 4 states are defined, 3 are used ('verify' isn't) and 5
-    # exist ('confirm' seems to have existed)
     state = fields.Selection(
         [
             ("draft", "Draft"),
@@ -257,6 +164,7 @@ class HrPayslip(models.Model):
         compute="_compute_payslip_count", string="Payslip Computation Details"
     )
     hide_child_lines = fields.Boolean(string="Hide Child Lines", default=False)
+    compute_date = fields.Date("Compute Date")
 
     @api.depends("line_ids", "hide_child_lines")
     def _compute_dynamic_filtered_payslip_lines(self):
@@ -336,9 +244,6 @@ class HrPayslip(models.Model):
             "context": {},
         }
 
-    def check_done(self):
-        return True
-
     def unlink(self):
         if any(self.filtered(lambda payslip: payslip.state not in ("draft", "cancel"))):
             raise UserError(
@@ -348,9 +253,6 @@ class HrPayslip(models.Model):
 
     def compute_sheet(self):
         for payslip in self:
-            number = payslip.number or self.env["ir.sequence"].next_by_code(
-                "salary.slip"
-            )
             # delete old payslip lines
             payslip.line_ids.unlink()
             # set the list of contract for which the rules have to be applied
@@ -362,11 +264,21 @@ class HrPayslip(models.Model):
                     date_from=payslip.date_from, date_to=payslip.date_to
                 ).ids
             )
-            lines = [
-                (0, 0, line)
-                for line in self._get_payslip_lines(contract_ids, payslip.id)
-            ]
-            payslip.write({"line_ids": lines, "number": number})
+            # get localdict and linesdict
+            localdict, lines_dict = self._get_payslip_lines(contract_ids, payslip.id)
+            # write payslip lines
+            number = payslip.number or self.env["ir.sequence"].next_by_code(
+                "salary.slip"
+            )
+            lines = [(0, 0, line) for line in list(lines_dict.values())]
+            payslip.write(
+                {
+                    "line_ids": lines,
+                    "number": number,
+                    "state": "verify",
+                    "compute_date": fields.Date.today(),
+                }
+            )
         return True
 
     @api.model
@@ -523,41 +435,7 @@ class HrPayslip(models.Model):
 
         return res
 
-    def get_baselocaldict(self, contracts):
-        """Basic dictionary values that are useful in most salary rules. Inherited
-        classes that overload this method should use the name of the module as
-        the dictionary key.
-
-        This method is evaluated once per payslip.
-
-        :param contracts: Recordset of all hr.contract records in this payslip
-        :return: a dictionary of discreet values and/or Browsable Objects
-        """
-        self.ensure_one()
-
-        worked_days_dict = {}
-        inputs_dict = {}
-        payslip = self
-        for worked_days_line in payslip.worked_days_line_ids:
-            worked_days_dict[worked_days_line.code] = worked_days_line
-        for input_line in payslip.input_line_ids:
-            inputs_dict[input_line.code] = input_line
-        inputs = InputLine(payslip.employee_id.id, inputs_dict, self.env)
-        worked_days = WorkedDays(payslip.employee_id.id, worked_days_dict, self.env)
-        payslips = Payslips(payslip.employee_id.id, payslip, self.env)
-        payroll_dict = BrowsableObject(
-            payslip.employee_id.id, payslip.get_payroll_dict(contracts), self.env
-        )
-
-        baselocaldict = {
-            "payslip": payslips,
-            "worked_days": worked_days,
-            "inputs": inputs,
-            "payroll": payroll_dict,
-        }
-        return baselocaldict
-
-    def get_contract_dict(self, contract, contracts):
+    def get_current_contract_dict(self, contract, contracts):
         """Contract dependent dictionary values.
         This method is called just before the salary rules are evaluated for
         contract.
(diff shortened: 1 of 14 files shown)


nimarosa:
Now i will patch hr_payslip_change_state module to support the new state functionality since it gives the test errors in Travis. Will add us as mantainers so we can modify this module in future versions. 

After this PR i will work in merging the cancel and change_state module so it's useful to be maintainers of them to be able to modify them.

ghost:
I haven't really tested the code, but I have studied it.
- (pr61) hr_payslip.py should have `_order = "id desc"`
- I think these lines may be deleted (they are also in [hr_salary_rule.py](https://github.com/OCA/payroll/blob/14.0/payroll/models/hr_salary_rule.py#L223)):
                localdict["result"] = None
                localdict["result_qty"] = 1.0
                localdict["result_rate"] = 100
                localdict["result_name"] = None

nimarosa:
@appstogrow 

Hello, 

1. Yes i overridden your change. Will change in following commit 
2. I don't quite understand. Can you elaborate more in this?

nimarosa:
@appstogrow I changed the order.

ghost:
I think that the 4 localdict-lines are not necesssary. But I didn't test it. So let's forget about that for now and approve this PR.

ghost (approved):

mtelahun:
Hi guys, can you give me a day to review this PR please? Thanks.

nimarosa:
@mtelahun yes no problem. We can wait till tomorrow and we can merge this and the other ones. 

Please also guys check the other PRs that brings the cancel and state module to base. The functionality is the same I just copied and adapted the functions but if you want to suggest more improvements to these functions will be great to do it now before the merge.

payroll/views/hr_payslip_views.xml
```diff
@@ -330,26 +334,45 @@
...
                 />
                 <field name="date_from" />
+                <field name="contract_id" />
+                <field name="payslip_run_id" />
                 <filter
-                    string="Draft"
+                    string="Pending Compute"
                     name="draft"
-                    domain="[('state','=','draft')]"
-                    help="Draft Slip"
```
mtelahun: I would prefer this stayed as "Draft". It's more intuitive and correctly reflects the state of the payslip. "Pending Compute" just sounds like a gratuitously round-about way of saying that. Also, it seems inaccurate because the next state is called 'Waiting'.

payroll/models/hr_payslip.py
```diff
@@ -275,6 +183,7 @@ def _compute_details_by_salary_rule_category(self):
                 "line_ids"
             ).filtered(lambda line: line.category_id and line.appears_on_payslip)
 
+    @api.depends("line_ids")
```
mtelahun: Is this really necessary? The field is not a stored field so it gets recomputed dynamically. Or should the field be stored?

nimarosa:
@mtelahun Did you have the opportunity to check this? So we can move forward

payroll/models/hr_payslip.py
```diff
@@ -362,11 +265,21 @@ def compute_sheet(self):
                     date_from=payslip.date_from, date_to=payslip.date_to
                 ).ids
             )
-            lines = [
-                (0, 0, line)
-                for line in self._get_payslip_lines(contract_ids, payslip.id)
-            ]
-            payslip.write({"line_ids": lines, "number": number})
+            # get localdict and linesdict
+            localdict, lines_dict = self._get_payslip_lines(contract_ids, payslip.id)
```
mtelahun: You've changed the signature of _get_payslip_lines() to return a two dictionaries, instead of a list, and one of the returned values (localdict) isn't even used. I **strongly** disagree with changing the signature of a method that other developers may be depending on in their modules. And after you've done that it appears you don't even use the newly returned value. I don't understand. What's the point?

payroll/models/hr_salary_rule_category.py
```diff
@@ -26,6 +26,15 @@ class HrSalaryRuleCategory(models.Model):
...
 
+    def _sum_salary_rule_category(self, localdict, amount):
+        self.ensure_one()
+        if self.parent_id:
+            localdict = self.parent_id._sum_salary_rule_category(localdict, amount)
+        localdict["categories"].dict[self.code] = (
+            localdict["categories"].dict.get(self.code, 0) + amount
+        )
+        return localdict
+
```
mtelahun: This method doesn't belong in hr.salary.rule.category. The only situation the method is used in is payslip calculation. There is no reason I can see to pollute the the hr.salary.rule.category object with knowledge of hr.payslip internal state (localdict, amount). It belongs in hr.payslip.

payroll/models/hr_payslip.py
```diff
@@ -587,117 +466,127 @@ def get_contract_dict(self, contract, contracts):
...
+                "payroll": BrowsableObject(
+                    self.employee_id.id, self.get_payroll_dict(contracts), self.env
+                ),
+                "current_contract": BrowsableObject(self.employee_id.id, {}, self.env),
+                "categories": BrowsableObject(self.employee_id.id, {}, self.env),
+                "rules": BrowsableObject(self.employee_id.id, {}, self.env),
+                "result_rules": BrowsableObject(self.employee_id.id, {}, self.env),
+            },
+        }
+        return localdict
```
mtelahun: The following keys either already have some way to be overridden or they are odoo models that can be directly modified by the developer:
"payslip", "worked_days", "inputs", "payroll", "current_contract"
The "categories"  and "rules" keys are there to be modified by the salary rule calculation code so if someone wants to mess with them the proper thing for them to do is inherit the salary calculation code.

You took what was a relatively straight-forward method of injecting arbitrary values into to be used **by** the salary rules and it seems you tried to turn it into a general mechanism for overriding the calculations themselves but in the process you made it overly complicated. 

Also, do you have a reason for making this change? And I don't mean a hypothetical "it may be useful in the future" reason, but an "I need it because I'm doing so and so and this is the only way I can thing of to achieve it' reason. If not then I would suggest to leave this piece of code alone.

payroll/models/hr_payslip.py
```diff
@@ -587,117 +466,127 @@ def get_contract_dict(self, contract, contracts):
...
+    def _get_baselocaldict(self, contracts):
+        self.ensure_one()
+        worked_days_dict = {
+            line.code: line for line in self.worked_days_line_ids if line.code
+        }
+        input_lines_dict = {
+            line.code: line for line in self.input_line_ids if line.code
+        }
+        localdict = {
+            **{
```
mtelahun: Sorry, I don't understand why this is necessary.

mtelahun (commented):
In general I found this to be a bit more than just a re-factoring and would have preferred it to be introduced in smaller chunks. I'm still trying to understand the full impact of it. Also, for such an extensive change in the payroll code you didn't provide any additional tests. I hate to sound like a broken record, but I'm a big believer in tests :grin:

nimarosa:
@mtelahun Hi. Yes I don't Provided tests because all functions are included in the current tests. 

Indeed the functionality is the same, that's why no extra tests are required from my point of view. 

The major changes are in the way the functions are structured and in their return values that don't affect the current functionality but allow inheriting the functions in external modules more easy and the new returning values allow to even modify the localdict directly in salary rules or custom modules. But the flow and functionality of the module remains intact that's why current tests continue passing.

The error in tests that show this PR is because BaseBrowsableObject and its child's are now in a new file, but in reality this clases are also tested currently. This error will be gone after merging because lines from payslip will be removed in favor of the new file. That's a common issue with codecov analysis. 

I'm not a test person, I focus on maintaining codecov green since I don't know which other tests can be made. Anyways, if you want to suggest some test or make que PR to my branch with your approach I'm happy to include it. 

Hope i clarified this a bit for you. 
From my point of view current tests are enough because no functionality is changed, only the functions to allow easy inheriting. 

I'm open to any suggestions.

payroll/views/hr_payslip_views.xml
nimarosa: Yes we can change that. I wanted to make filters more user friendly but maybe it's not the correct words. Will change next commit.

payroll/models/hr_payslip.py
nimarosa: The point of returning localdict is that when you inherit this method, you can't for example add a functionality that requires localdict because the function always return the lines. 

For example this is useful if you have another type of categories in a private module and you need to don some method like _sum_by_salary_rule_category. 

With the current implementation you can't access localdict inheriting any function and that limits the developer to interact and change available values in localdict. 

It's no used because it's meant to be inherited, for example Can be used in @appstogrow PR or in any implementation requiring to modify localdict on the middle of calculation. 

Hope it's clear if not I can submit que example in the code.

payroll/models/hr_salary_rule_category.py
nimarosa: This function always used localdict and amount as params if that is what you are saying. 

The purpose to include it there is because you can need accumulate the categories outside the payslip calculation. 

For example I have a module that makes a report based on accumulated values in categories. If I only can see them in the payslip calculation I have to create a identical method which might be writing again code that is already in the module. 

Also allows to modify this calculation function without the need to rewrite all the get payslip lines function to inject something there. 

My current known about object oriented coding also tells me that this function belong there because it's a method of the category, the category itself offers that functionality. But it think that is more a common dev discussion haha. 

I'm more into of separating concerns in the objects they belong. 

I think it makes no harm to have it there but I could be wrong. What's true is that this function need to be separated from the loop in order to be inherited and modified. 

If you want to modify how categories are calculated now it's easy, before it was rewriting all the function.

payroll/models/hr_payslip.py
nimarosa: Actually the reason for it is that you can override them without touching the calculation loop which is what i intended to do. 

I use this for example to use localdict context in other places like a module that simulates payslip calculation and actually in a module for profit calculation that needs to inject values in the localdict to be summed with current calculation values and then writed to a table. 

With current behavior if you want to create or edit dicts inside the salary rules, you can do it (it's supported) but you then don't have a way to take that values when the calculation loop ends. 

That's why localdict is also a returned value, because I have already two implementations that require doing calculations on the fly when the rules are being calculated. If we only return the lines, we can't get any other data out the loop. For easy payroll structured this is okay but at least in the implementations that i had to work on, we need to update tables on the fly to get updated values that are dependencies of other rules. 

I know it seems complicated but I've seen code like this in other modules of odoo. 

I invite you to try it and you will see it opens a new whole of new possibilities to allow to inject values, and also it makes no harm because they still are all private functions. 

Then I can show you in my private repo if you want to see how this is useful, really, I think you would like it because it's very similar to your payroll and current_contract dict approach but allows a new level of customization to happen in salary rules.

payroll/models/hr_payslip.py
nimarosa: This is useful if you want to merge the dicts and be available as keyword arguments. It's similar to **kargs in functions.

nimarosa: It can be removed, I added it because it makes no change an allow to inject a whole dict easily if needed but this yes I don't use it for now. 

But if not used it makes no difference.

payroll/models/hr_payslip.py
nimarosa: It's true is not necesarry, will change it next commit.

mtelahun:
But if you want to get at localdict you can inherit 
_compute_payslip_line() now, right?

This PR changes the signature of _get_payslip_lines(), which will affect 
all modules which have been inheriting it up to now.

mtelahun:
localdict contains information about the internal state of hr.payslip, 
which hr.salary.rule.category should not have access to.  Presumably, 
the report module you mentioned is reporting on payslips (I assume) so 
it is already manipulating/accessing hr.payslip. That's why I said this 
method belongs there.

If you need hr.salary.rule.category to provide a method that recursively 
adds amounts related to category codes then you should abstract away 
only that part of it and it should be a public method. For example:

```

def sum_salary_rule_category(self, categories_dict, amount):

     ...

     categories_dict[self.code] = categories_dict.get(self.code, 0) + amount

     ...

```

mtelahun:
If there is no current need for this I prefer it be removed. If it ever 
becomes necessary in the future it can be added then.

mtelahun:
On 07/09/2022 00:02, Nicolas Rodriguez Sande wrote:
> Actually the reason for it is that you can override them without 
> touching the calculation loop which is what i intended to do. 

If a module needs to change one of these values in the middle of rule 
evaluation then inheriting _compute_payslip_line() should be sufficient.


> I use this for example to use localdict context in other places like a 
> module that simulates payslip calculation and actually in a module for 
> profit calculation that needs to inject values in the localdict to be 
> summed with current calculation values and then writed to a table.
>
> With current behavior if you want to create or edit dicts inside the 
> salary rules, you can do it (it's supported) but you then don't have a 
> way to take that values when the calculation loop ends.
>
I don't think OCA/Payroll should be encouraging people to modify it's 
payslip calculation internal state from salary rules. That is a recipe 
for disaster. I just think that's so wrong on so many levels. I think 
this is a functionality that belongs in a private repo and not in OCA. 
If it is already possible that is a bug, not a feature IMO. If your 
concern is getting the state of localdict after the payslip calculation 
ends (or anywhere in between) the simplest thing to do is call a hook 
method and pass it localdict instead of all these complications.

```

def _get_payslip_lines(....):

     localdict = ...

     for contract in contracts:

         ...

         for rule in rules:

             ...

     self.localdict_hook(localdict)

     return lines

```


> That's why localdict is also a returned value, because I have already 
> two implementations that require doing calculations on the fly when 
> the rules are being calculated. If we only return the lines, we can't 
> get any other data out the loop. For easy payroll structured this is 
> okay but at least in the implementations that i had to work on, we 
> need to update tables on the fly to get updated values that are 
> dependencies of other rules.
>
I would be hard-pressed to think of a situation that can't be addressed 
with the current functionality. I'm not saying it's impossible, but I 
would need an example to convince me.


> I know it seems complicated but I've seen code like this in other 
> modules of odoo.
>
> I invite you to try it and you will see it opens a new whole of new 
> possibilities to allow to inject values, and also it makes no harm 
> because they still are all private functions.Then I can show you in my 
> private repo if you want to see how this is useful, really, I think 
> you would like it because it's very similar to your payroll and 
> current_contract dict approach but allows a new level of customization 
> to happen in salary rules.
>
If you could provide a concrete example that would help me understand 
what this part of the PR is meant to achieve.

ghost:
@nimarosa 
1. Are you able to do your implementations if _get_payslip_lines() will continue to return a list of lines? For example by overriding compute_payslip_line() or adding a hook inside _get_payslip_lines().
2. What do you think about moving _sum_salary_rule_category() back to hr.payslip? Would it be useful for you to have this method in hr.salary.rule.category without localdict?

ghost:
**baselocaldict**
AFAIK, the purpose of **localdict** is to provide variables for the code in salary rules.
In `condition_python` and `amount_python_compute`, the default values mention these variables:
- payslip (baselocaldict)
- employee (added in _get_payslip_lines() inside contracts loop)
- contract (added in _get_payslip_lines() inside contracts loop)
- rules (added in _get_payslip_lines() before contracts loop; this PR moves it to baselocaldict)
- categories (added in _get_payslip_lines() before contracts loop; this PR moves it to baselocaldict)
- worked_days (baselocaldict)
- inputs (baselocaldict)
- payroll (baselocaldict)
- current_contract (added in _get_payslip_lines() before contracts loop; this PR moves it to baselocaldict)

> You took what was a relatively straight-forward method of injecting arbitrary values into to be used by the salary rules and it seems you tried to turn it into a general mechanism for overriding the calculations themselves but in the process you made it overly complicated.

@mtelahun 
1. Does **baselocaldict** have other purposes than providing variables for salary rules?
2. Why are "payslip", "worked_days" and "inputs" part of baselocaldict, instead of being added inside `_get_payslip_lines()`?
3. Does it not make sense to add "rules", "categories" and "current_contract" together with "payslip", "worked_days" and "inputs"?

ghost:
@nimarosa
I see that you add `result_rules` to baselocaldict and that you add some values to it inside `_compute_payslip_line()`. But it is not used to create the payslip line. I suggest that you remove this code and implement it in a custom module by overriding _compute_payslip_line(). I also need to override this method and do many changes. I suggest that for now we make the generic _compute_payslip_line() as simple as possible. Later we can compare our custom implementations and possibly suggest improvements for _compute_payslip_line().

nimarosa:
@appstogrow @mtelahun Okay I will try to answer in parts because to many information. 

- About salary rule category
If the method is in categories file or in payslip file it's irrelevant for me. The only thing I will like to maintain is that it has to be a separated method. Currently if you look at the code, the sum category method is defined inside _get_payslip_lines function, which for me is bad because if you want to inherit the way that categories are calculated you have to override the full method because it's defined inside. 
Are you okay to having this method as a private method in hr_payslip? For me it's not important that it has to be in the categories file but to be a separate method not defined inside a function (which I think is a bad practice to have method definitions inside functions).

- About localdict return in _get_payslip_lines. 
You are saying that if I need localdict I can inherit the new _compute_payslip_line method. Yes it's true but that's if I need localdict on each line calculation of the loop (its useful for some cases). But what if you need to access localdict when the calculation loop has finished, meaning when the compute function has finished running and then you need to get the localdict and do something with values inside it? 
That's why localdict is returned in compute payslip lines and in get payslip lines. 

I will look the hook approach more in depth today, if this can solve this problem, I change it. The only thing I need and I think is useful is having the possibility to access localdict object in every step possible. 

@mtelahun im happy to provide you an example, wait for me to get to the office and I copy a bit of code of the usage of this. 

In the meantime we continue discussing the approach, like I said if there is a better way of doing it, I'm open to it.

nimarosa:
> **baselocaldict**
> 
> AFAIK, the purpose of **localdict** is to provide variables for the code in salary rules.
> 
> In `condition_python` and `amount_python_compute`, the default values mention these variables:
> 
> - payslip (baselocaldict)
> 
> - employee (added in _get_payslip_lines() inside contracts loop)
> 
> - contract (added in _get_payslip_lines() inside contracts loop)
> 
> - rules (added in _get_payslip_lines() before contracts loop; this PR moves it to baselocaldict)
> 
> - categories (added in _get_payslip_lines() before contracts loop; this PR moves it to baselocaldict)
> 
> - worked_days (baselocaldict)
> 
> - inputs (baselocaldict)
> 
> - payroll (baselocaldict)
> 
> - current_contract (added in _get_payslip_lines() before contracts loop; this PR moves it to baselocaldict)
> 
> 
> 
> > You took what was a relatively straight-forward method of injecting arbitrary values into to be used by the salary rules and it seems you tried to turn it into a general mechanism for overriding the calculations themselves but in the process you made it overly complicated.
> 
> 
> 
> @mtelahun 
> 
> 1. Does **baselocaldict** have other purposes than providing variables for salary rules?
> 
> 2. Why are "payslip", "worked_days" and "inputs" part of baselocaldict, instead of being added inside `_get_payslip_lines()`?
> 
> 3. Does it not make sense to add "rules", "categories" and "current_contract" together with "payslip", "worked_days" and "inputs"?

Like Henrik said, 

In localdict, the only change I made is that before this PR the _get_baselocaldict only returns the values for worked days, inputs, payroll. Then in _get_payslip_lines the old approach adds to localdict empty dicts on the beginning for the categories, rules, etc. 

The only reason and change that this introduce is that we now have a function that returns the default localdict and that function is called in get payslip lines. Before we have localdict manipulation spread all over the get payslip lines function which made the function bloated and less readable. 

In resume: this method don't changes anything of how it was done before. It only centralices the localdict object definition in only one function instead of having it for all over the place. 

Please check the current get baselocaldict  function and then the beginning of the get payslip lines function and you will see what I'm talking about.

nimarosa:
> @nimarosa I see that you add `result_rules` to baselocaldict and that you add some values to it inside `_compute_payslip_line()`. But it is not used to create the payslip line. I suggest that you remove this code and implement it in a custom module by overriding _compute_payslip_line(). I also need to override this method and do many changes. I suggest that for now we make the generic _compute_payslip_line() as simple as possible. Later we can compare our custom implementations and possibly suggest improvements for _compute_payslip_line().

@appstogrow Actually result_rules is a new object that can be used in salary rule computation. It allows you to access all rule values (qty, rate, amount, and total). With current implementation you can only access the total rule value. I find useful to access all line values for example in lines that require to know the quantity of the previous one, or even lines that need to know the unitary_value (amount) and not the total. 

This is something i saw it's possible in enterprise so i replicated the functionality in our code. 

I will add documentation about this, because it has to be in the available objects list in salary_rules. 

-----
Keep posted i will do a few commits today to try to implement the changes discussed.

mtelahun:
Hi @nimarosa,


On 07/09/2022 15:11, Nicolas Rodriguez Sande wrote:
>
> @appstogrow <https://github.com/appstogrow> @mtelahun 
> <https://github.com/mtelahun> Okay I will try to answer in parts 
> because to many information.
>
>  *
>
>     About salary rule category
>     If the method is in categories file or in payslip file it's
>     irrelevant for me. The only thing I will like to maintain is that
>     it has to be a separated method. Currently if you look at the
>     code, the sum category method is defined inside _get_payslip_lines
>     function, which for me is bad because if you want to inherit the
>     way that categories are calculated you have to override the full
>     method because it's defined inside.
>     Are you okay to having this method as a private method in
>     hr_payslip? For me it's not important that it has to be in the
>     categories file but to be a separate method not defined inside a
>     function (which I think is a bad practice to have method
>     definitions inside functions).
>
  That's works for me. Thanks.


>
>  *
>
>     About localdict return in _get_payslip_lines.
>     You are saying that if I need localdict I can inherit the new
>     _compute_payslip_line method. Yes it's true but that's if I need
>     localdict on each line calculation of the loop (its useful for
>     some cases). But what if you need to access localdict when the
>     calculation loop has finished, meaning when the compute function
>     has finished running and then you need to get the localdict and do
>     something with values inside it?
>     That's why localdict is returned in compute payslip lines and in
>     get payslip lines.
>
That's why I suggested the hook method. My main concern is that 
_get_payslip_lines() return value is being changed, which will impact 
every developer who has inherited that method. The hook method avoids 
that while giving you the functionality that you need.

mtelahun:
Ok. Let me take a look at it again.

mtelahun:
On 07/09/2022 16:46, Nicolas Rodriguez Sande wrote:
>
> I will add documentation about this, because it has to be in the 
> available objects list in salary_rules.
>
> ------------------------------------------------------------------------
>
> Keep posted i will do a few commits today to try to implement the 
> changes discussed.
>
Hi Nicolas,

Can I make a suggestion? If you're interested you can do it and if not 
that's fine also.

There are quite a few bits of this PR that are really interesting and 
that I think can be merged right away with very little discussion if you 
create separate PRs for them.

1. The "verify" state and the modifications to the other modules to take 
advantage of it.

2. The refactoring into _compute_payslip_line()

3. Move out _sum_category_rules() into its own function. As I said my 
only concern with this function is that if it continues to access 
localdict is should be in hr.payslip. But if it is in 
hr.salary.rule.category it should be abstracted out a bit so that you 
only pass categories dict and not the whole localdict.


This allows most of the functionality in this PR to be merged right away 
while we discuss issue of localdict and baselocaldict.

nimarosa:
@mtelahun Hello, actually I tried the hook method approach and I think it works great for the functionality that I was trying to achieve and also i have to recognize that its also a more cleaner solution and like you said we skip having to change _get_payslip_lines return values so we don't mess with modules which can be inheriting it. 

So in next commits I will make this change and the other ones discussed here (sum category function, filter and little more tweaks) 

You are right maybe those little changes would be better in another PR but since I rebased the branch now everything is in only one commit and it will be more difficult to extract it now. 

Since I'm adopting your sugestión to do a hook function, I think that with new changes we will all of us agree to merge this completely. 

I will submit the new commits after I finish a meeting and we can do a final review of everything.

nimarosa:
@mtelahun @appstogrow 

Okay guys, i changed the _sum_salary_rule_category method to payslip. I added "category" as a parameter since if this method is in payslip we need to pass the rule category for allow the sum to happen. Before it was not necessary because it was called directly from the category object. But it's working now. 

I also, removed the localdict as return value so now the function _get_payslip_lines returns lines_dict as always. And we have a new localdict_hook method, that allow access to the localdict when the calculation loop is finished. Also this hook overrides localdict in the end of the function, in case we want to inherit _get_payslip_lines and do something with the hooked localdict. But the function is returning as always which was the main requirement. I also liked this approach more than mine so i think we have so quorum here. 

Do you have any more comments? If not, i think we can merge when tests finish passing. 

Do you think we should rebase everything in one commit?

nimarosa:
> I haven't really tested the code, but I have studied it.
> 
> * (pr61) hr_payslip.py should have `_order = "id desc"`
> * I think these lines may be deleted (they are also in [hr_salary_rule.py](https://github.com/OCA/payroll/blob/14.0/payroll/models/hr_salary_rule.py#L223)):
>   localdict["result"] = None
>   localdict["result_qty"] = 1.0
>   localdict["result_rate"] = 100
>   localdict["result_name"] = None

@appstogrow About that, i tested it and these lines are necessary. If we dont default this values in the loop the rule takes the values of the last rule. Leaving at it is for now.

ghost:
> result_rules: object with a dict of qty, rate, amount an total of calculated rules

Would the dict of qty, rate, amount and total afterwards be like the totals here in the bottom? I don't really see the point. (I added these totals in a test database to illustrate.)
image

BTW
This total is in the payslip view, and it doesn't make sense when using leaves_positive, which was introduced in #50.
image

payroll/models/hr_payslip.py
```diff
@@ -587,117 +468,135 @@ def get_contract_dict(self, contract, contracts):
...
```
mtelahun: If you currently don't need this I would prefer you leave it out. If you use it in the future it can always be re-introduced.

mtelahun (approved):

nimarosa:
> > result_rules: object with a dict of qty, rate, amount an total of calculated rules
> 
> 
> 
> Would the dict of qty, rate, amount and total afterwards be like the totals here in the bottom? I don't really see the point. (I added these totals in a test database to illustrate.)
> 
> image
> 

Hi, result_rules creates one dict item per salary rule. So it's not a sum of all lines. To illustrate you can now do (inside salary rules): 

result_rules.BASIC.qty 
result_rules.BASIC.rate
result_rules.BASIC.amount 
result_rules.BASIC.total

As I said earlier with current implementation, in a salary rule you can only access the previously calculated rules **totals**, so if you need any other data of previous calculated rules like the amount of a line or the rate used in other rule, currently you can't take that data. 
With this new dict you have available in salary rules all lines data if you need it for calculations. 


> > BTW
> 
> This total is in the payslip view, and it doesn't make sense when using leaves_positive, which was introduced in #50.
> 
> image
> 

Yes we can remove it in this PR if you want. It only makes sense if you use the leaves negative option. Will change it if you want because I think we can't have a condition to show it depending of setting, no?

payroll/models/hr_payslip.py
nimarosa: Actually I forgot to tell you. I tried removing the ** and if I do then I get an error because a dict is not accesible. 

I think I put that there because of that. Take into account that the dict is inside another dict and that's why ** allow merging it with the parent dict and make the values accessible. 

I can try to remove it completely and only leave one dict but en last commit I tested without it and all tests fail. 

Will try getting it out in a while and if not we will have to leave it there.

nimarosa:
@mtelahun I've just removed the ** from baselocadict. This is done. 

@appstogrow 1. I removed the sum in the worked_days tables. If then we think of a way to having it, we add it. For now it's removed. 2. I leaved the result_rules dict, because is useful at least for some complex salary rules that i have, previously of this change to get those values i have to inherit functions to get the values i wanted, this way i have it in a dict. I also added BaseBrowsableObject to this method so it can be accessed from salary rules with dot notation. 

Please review but i think we can merge this and continue with hr_payslip_cancel and hr_payslip_change_state merge PRs. 

Wait for your comments.

mtelahun:
Doesn't removing the outer {} fix it?

nimarosa:
> Doesn't removing the outer {} fix it?

Yes I've done it in last commits.

ghost:
Thank you @nimarosa 
I approve this PR.

nimarosa:
@mtelahun are you okay with this? If yes I merge now and we can continue to approve the others so Henrik can start working in his PR

nimarosa:
I didn't see that @mtelahun has already approved the changes 11 hours ago and since i changed his last suggestion i think we are okay to merge. 

I will merge this now and rebase the other PRs to continue merging. Next week i have a vacation and i want to let this module the more upgraded possible so other people can continue working on it. 

Merging. 

/ocabot merge minor

Issue #54, opened 2022-08-16
Merge of hr_payroll_cancel and hr_payroll_change_state. 


nimarosa:
Modules hr_payroll_cancel and hr_payroll_change_state. 

Should we merge this modules logic into payroll module itself? Or maybe just provide this functionality in the base payroll module so we don't need that modules? 

I don't think the point of maintaining more two modules whose purpose  is to add the functionality of mass change states and cancelation. We can just plug-in this functionality to the base module. 

What do you think?

mtelahun:
While the functionality in these modules is useful it's not essential 
for the payroll module to function correctly. I think we should keep the 
base payroll module as basic as possible and allow people to pick and 
choose what other functionality they want by installing other modules on 
top of it.

pedrobaeza:
I consider the cancel option a base feature.

mtelahun:
I disagree, but I won't oppose the merge.

pedrobaeza:
MIchael, take into account that anyone may make any mistake. and not being able to amend it is not good IMO.

mtelahun:
Hi Pedro,

Like I said I won't oppose the change. I also understand that anyone can 
make a mistake and in that scenario the ability to cancel the payslip 
and start again is useful. But just for arguments sake consider the 
following scenario. Let's say you're a company that has tens of 
thousands of employees and you have an automated payroll system. In this 
scenario you don't want to give your payroll staff the ability to modify 
individual payslips. For one thing, it's inefficient and second, with 
that many employees it becomes harder to detect fraud. If a payroll 
staff notices a problem you want them to change the **source** of the 
problem and then re-generate the payslip through the system.

I understand that a lot of Odoo users are of the small and medium 
enterprise type that do not have the problems I'm talking about and 
would prefer to have the ability to cancel a payslip built-in rather 
than install an additional module. Also, if you have tens of thousands 
of employees you should also have the ability to install or build a 
module that takes that functionality away. So it could be argued that a 
majority of users prefer to have the functionality, which is why I won't 
oppose the change.

I don't take opposing views just to be confrontational I usually have a 
reason. I hope you don't mind :-P

pedrobaeza:
Thanks for bringing such concerns. I think we can just protect that feature under a configuration option "Enable payslip cancelling"  or similar, so that only system administrators can modify this and if disabled, payroll managers won't be able to do it although the feature is integrated. What do you think?

nimarosa:
[...]

PR #67, opened 2022-09-08
[IMP] payroll: Merge payslip_cancel and payslip_change_state in base module

nimarosa:
I've made a new PR since rebasing the others was unsuccessful haha. 

- As discussed in https://github.com/OCA/payroll/issues/54 here we are merging the cancel functionality to the base module and removing the hr_payslip_cancel module. As discussed, the functionality is activated with a setting, that allows or not to cancel confirmed payslips to the user.
Everything is working the only thing i couldn't think of is hiding the "Cancel Payslip" button when the setting is not True.
For now the button is shown in draft,verify and done state. If anyone has a suggestion on how to hide this button in state "done" when the setting is not activated, will include it here.

- As discussed in https://github.com/OCA/payroll/issues/54 here we are merging the change state wizard to the base module and removing the hr_payslip_change_state module. If someone want to suggest changes to the module or improve the functionality this is a good change to do it, so we can merge a more improved functionality.

Regards.


nimarosa:
@apps2grow @mtelahun 

I've made a new PR for merging this modules since the rebase was not simple as i thought. 

Okay will see if tests passes. It's pending to write tests for hr_payslip_cancel since the tests in the previous model now have to take into account the setting. Will try to do it today, if someone want to help with tests and also add suggestions to this merges, feel free to make PR to my branch.

payroll/tests/test_hr_payroll_cancel.py
```diff
@@ -10,6 +10,12 @@
 
 class TestHrPayrollCancel(common.TransactionCase):
     def setUp(self):
+
+        # Set system parameter
+        self.env["ir.config_parameter"].sudo().set_param(
+            "payroll.allow_cancel_payslips", True
+        )
+
         super(TestHrPayrollCancel, self).setUp()
```
mtelahun: This super() should be the first thing you call. That's why the test is failing.

payroll/views/hr_payslip_views.xml
```diff
@@ -109,7 +109,7 @@
                         string="Cancel Payslip"
                         name="action_payslip_cancel"
                         type="object"
-                        states="draft,confirm,verify"
+                        states="draft,confirm,verify,done"
```
mtelahun: What you can probably do here is
1. create a computed field, allow_payslip_cancel, in hr.payslip that mirrors the value of the "cancel payslip" setting
2. leave this button as is and create another Cancel button which is only shown in Done state add an attribute to the effect of: 'invisible': [('allow_payslip_cancel', '=', False)]

payroll/models/hr_payslip.py
```diff
@@ -211,8 +214,23 @@ def action_payslip_done(self):
...
+                if payslip.refunded_id and payslip.refunded_id.state != "cancel":
+                    raise ValidationError(
+                        _(
+                            """To cancel the Original Payslip the
+                        Refunded Payslip needs to be canceled first!"""
+                        )
+                    )
+            else:
+                if self.filtered(lambda slip: slip.state == "done"):
+                    raise UserError(_("Cannot cancel a payslip that is done."))
```
mtelahun: Does the current behaviour allow you to cancel a payslip that is not in 'Done' state? If so, then I would suggest checking 'allow_cancel_payslips' only if the state is in done.

mtelahun (approved):
LGTM code review

payroll/models/hr_payslip.py
nimarosa: Yes the payslip can be canceled in any state. When the setting is False, payslip in done state can not be canceled. When it's true, they can be canceled.

payroll/views/hr_payslip_views.xml
nimarosa: Will do this next commit. Thanks for the idea.

nimarosa:
@mtelahun Can you get your help here with the tests? I hate test so much. The test is the same that was in the module, i don't know why it don't find the journal_id, it's created before that line that fails. 

I already made your suggested changes, everything perfect.

ghost:
`payroll` does not depend on `account`.
Try to remove account_debit, account_credit, account_journal etc. from the tests.

ghost:
I see that `hr_payslip_change_state` depends on `hr_payroll_cancel`, which depends on `payroll_account`.
So I think that these modules should be merged into **payroll_account**, not `payroll`.

**EDIT**
The word **account** does not otherwise exist in `hr_payroll_cancel `or `hr_payslip_change_state`. So please continue merging with `payroll`, and remove account stuff from the tests.

nimarosa:
@appstogrow yes there is a piece of code that use account, if you see in the code I also edited the payroll_account function. 

The dependency in hr_payroll_cancel was there because in the module they inherited the method and added accounts to the calculation, now that I separated the function in two, we can remove the dependency. 

Will do it in next commits thanks for your help!

nimarosa:
Okay, tests are passing now. 
For the future, it should be ready to add some lines in payroll_account tests to test the cancelation, but the functionality is very simple so we can merge this way.
Please review and i think we can merge it.

ghost:
I see 2 Cancel Payslip buttons. Please change the following:
First button: states="draft,verify"
Second button: attrs="{'invisible': ['|', ('state', '!=', 'done'), ('allow_cancel_payslips', '=', False)]}"

I think change state works as expected.

nimarosa:
@appstogrow Done. Can you check? I still see two buttons after that change but maybe is on my end. It should be appearing only one.

mtelahun (approved):
LGTM functional test.

ghost (approved):

PR #68, opened 2022-09-13
[FIX] [14.0] payroll: fix regression in #62

mtelahun:
When calling _compute_payslip_line() call it using the payroll object instead of recordset 'self'. If called on 'self' with multiple payslips in the recordset it triggers the 'Expecting singleton' exception.


nimarosa (approved):

nimarosa:
You are right. 

Merging. 

/ocabot merge patch

PR #69, opened 2022-09-17
[14.0][IMP] payroll: refactor many methods to work on recordsets. No functional changes.

mtelahun:
The payroll code has many vestiges of the pre- recordset API lying around. Many functions require you to pass in a payslip_id or contracts, etc when they can just access it from the current record. This is inefficient and can also lead to bugs (like the one fixed in #68). So, I took a swing at correcting many of those places.

The refactored methods are:
_get_employee_contracts()
_get_salary_rules()
onchange_struct_id()
onchange_dates()
onchange_employee()

I also refactored the functionality in _get_payslip_lines(). However, since we can't change its signature at this late date I put the refactored code in a new method: get_lines_dict() and made _get_payslip_lines() call it, instead. To make sure people start using the new method I put a deprecation warning whenever _get_payslip_lines() is called.

If the diff is too big to read at once try looking at the diffs for the individual commits. That should make it easier to digest. As mentioned in the title: There should be no functional changes introduced.

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -516,16 +516,22 @@ class HrPayslip(models.Model):
         }
         return localdict
 
-    def _get_salary_rules(self, contracts, payslip):
-        if len(contracts) == 1 and payslip.struct_id:
-            structure_ids = list(set(payslip.struct_id._get_parent_structure().ids))
-        else:
-            structure_ids = contracts.get_all_structures()
-        rule_ids = (
-            self.env["hr.payroll.structure"].browse(structure_ids).get_all_rules()
-        )
-        sorted_rule_ids = [id for id, sequence in sorted(rule_ids, key=lambda x: x[1])]
-        sorted_rules = self.env["hr.salary.rule"].browse(sorted_rule_ids)
+    def _get_salary_rules(self):
+        rule_obj = self.env["hr.salary.rule"]
+        sorted_rules = rule_obj
+        for payslip in self:
+            contracts = payslip._get_employee_contracts()
+            if len(contracts) == 1 and payslip.struct_id:
+                structure_ids = payslip.struct_id.ids
+            else:
+                structure_ids = contracts.get_all_structures()
+            rule_ids = (
+                self.env["hr.payroll.structure"].browse(structure_ids).get_all_rules()
+            )
+            sorted_rule_ids = [
+                id for id, sequence in sorted(rule_ids, key=lambda x: x[1])
+            ]
+            sorted_rules |= rule_obj.browse(sorted_rule_ids)
         return sorted_rules
 
     def _compute_payslip_line(self, rule, localdict, lines_dict):
@@ -592,7 +598,7 @@ class HrPayslip(models.Model):
             localdict = dict(
                 baselocaldict, employee=contract.employee_id, contract=contract
             )
-            for rule in self._get_salary_rules(contracts, payslip):
+            for rule in self._get_salary_rules():
                 localdict["result"] = None
                 localdict["result_qty"] = 1.0
                 localdict["result_rate"] = 100
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -683,11 +683,12 @@ class HrPayslip(models.Model):
         return localdict
 
     def _get_employee_contracts(self):
-        return self.env["hr.contract"].browse(
-            self.employee_id._get_contracts(
-                date_from=self.date_from, date_to=self.date_to
-            ).ids
-        )
+        contracts = self.env["hr.contract"]
+        for payslip in self:
+            contracts |= payslip.employee_id._get_contracts(
+                date_from=payslip.date_from, date_to=payslip.date_to
+            )
+        return contracts
 
     @api.onchange("struct_id")
     def onchange_struct_id(self):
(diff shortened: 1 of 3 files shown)


nimarosa:
@mtelahun looks very promising, we are doing a good refactoring. 

Yes the contract and payslip calls is something I wanted to get rid, great you've done it. 

Will review it in a while and make some suggestions I saw looking at code in my phone. 

Thanks for the new PR.

payroll/models/hr_payslip.py
```diff
@@ -281,25 +284,11 @@ def compute_sheet(self):
...
             number = payslip.number or self.env["ir.sequence"].next_by_code(
                 "salary.slip"
             )
-            lines = [
-                (0, 0, line)
-                for line in list(
-                    self._get_payslip_lines(contract_ids, payslip.id).values()
-                )
-            ]
+            lines = [(0, 0, line) for line in list(payslip.get_lines_dict().values())]
```
nimarosa, line 291: Since we are changing the function that is called here. Is there any reason to mantain _get_payslip_lines() since it will never be called. Correct me if I'm wrong.

payroll/models/hr_payslip.py
```diff
@@ -575,44 +570,54 @@ def _compute_payslip_line(self, rule, localdict, lines_dict):
...
+                            rule, localdict, lines_dict
+                        )
+                    else:
+                        # blacklist this rule and its children
+                        blacklist += [
+                            id for id, seq in rule._recursive_search_of_rules()
+                        ]
+                    lines_dict.update(_dict)
+                    # call localdict_hook
+                    localdict = payslip.localdict_hook(localdict)
```
nimarosa: Please note that localdict_hook was outside the loop in the previous implementation. 

For my implementations I need it outside the loop. But you gave me the idea to have two hooks, one inside the loop and one outside it. 

What do you think? They could be named: after_localdict_hook and onloop_localdict_hook. I'm open to better names haha.

payroll/models/hr_payslip.py
```diff
@@ -575,44 +570,54 @@ def _compute_payslip_line(self, rule, localdict, lines_dict):
...
+                # reset "current_contract" dict
+                baselocaldict["current_contract"] = {}
         return lines_dict
 
-    def localdict_hook(self, localdict, payslip):
```
nimarosa, line 614: I think passing "payslip" here is actually useful because when you inherit the hook you would probably need some payslip value. Payslip will we available on self if I inherit this? If yes, we can remove it. If not, I need payslip object there. But maybe it's available on self.

payroll/models/hr_payslip.py
```diff
@@ -683,61 +688,69 @@ def _sum_salary_rule_category(self, localdict, category, amount):
...
 
     def _get_employee_contracts(self):
-        return self.env["hr.contract"].browse(
-            self.employee_id._get_contracts(
-                date_from=self.date_from, date_to=self.date_to
-            ).ids
-        )
+        contracts = self.env["hr.contract"]
+        for payslip in self:
+            contracts |= payslip.employee_id._get_contracts(
```
nimarosa: Why is the |= needed here?

nimarosa (changes requested):
Please check comments. I didn't test it functionally but I've done a quick review of the code. Will do another one later but this is what I see in a quick view. 

Edit: please also check Travis failing.

payroll/models/hr_payslip.py
mtelahun: because self is a recordset which may include more than one payslip. When the method returns it will return all the contracts from all the payslips in the recordset, minus duplicates.

payroll/models/hr_payslip.py
mtelahun: Ahh yes, sorry. This is a mistake on my part. As for having both the inner and outer hook I'm don't know how much more useful it is than just inheriting _compute_payslip_line().

payroll/models/hr_payslip.py
mtelahun: You are right, but _get_pasylip_lines() has been around for a while and other modules most likely have inherited it. So, I don't want to break them. However, I've included a warning whenever they call it so that can update their code.

payroll/models/hr_payslip.py
mtelahun: Yes, the payslip is available as self. You can access the payslip values as self.xxxx, self.yyy.

I was going to add a self.ensure_one() call here, but I wasn't sure how you would use it so I didn't add it.

mtelahun:
BTW @apps2grow  I had an interesting use-case for git bisect on this PR. I initially had 6 commits that all passed payroll tests, but I hadn't tested them with payroll_account. When I created this PR ocabot caught the failing payroll_account test. But I had 6 commits! Which one could be the cause? I used git bisect and found the commit that caused the payroll_account test to fail. It took me less than 5 minutes. Long story short: if you can't break one large commit into different PRs at least break up your changes into small logical pieces, and commit them separately in the same PR. It will save you a lot of debugging!

nimarosa (approved):
Looks good now. @apps2grow do you want to review? 

Tomorrow I do functional test but I think it will be no problem. If not, I merge this.

Thanks @mtelahun for the contribution.

ghost:
I did a functional test. Creating and computing a payslip will call all 5 methods. No errors. :)

ghost (approved):

PR #70, opened 2022-09-19
[14.0][IMP] payroll: Compute rule / payslip line

ghost:


ghost:
This PR has 5 commits. I have tried to isolate each change, inspired by @mtelahun.

The methods below are split so that I can inherit `payroll` (in another module) and let a rule create **multiple payslip lines**.

* Rule: `_compute_rule_code()` may return a list of dicts for multiple lines. I will override `_get_rule_dict()` to handle this.
* Payslip: `_compute_rule()` may receive a list of dicts for multiple lines. I will override `_get_line_dict()` to handle this.

Again, I have experimented with changing a method name. I can change `_compute_rule()` back to `_compute_payslip_line()` if you think this is a better naming.

nimarosa:
Hello, will review it when i get home. About the names, i prefer not to change it to _compute_rule() since it's confusing with the other method from salary_rules also called compute_rule. 
But will put that in the review so we see a final name, since i have modules inheriting this.

payroll/models/hr_payslip.py
```diff
@@ -523,29 +523,33 @@ def _get_salary_rules(self):
             sorted_rules |= rule_obj.browse(sorted_rule_ids)
         return sorted_rules
 
-    def _compute_payslip_line(self, rule, localdict, lines_dict):
+    def _compute_rule(self, rule, localdict, lines_dict):
```
nimarosa: I think we should maintain `_compute_payslip_line` or if we change it, it should be different because `_compute_rule` is a method in salary rules. 

Do you find `_compute_payslip_line` not accurate?

payroll/models/hr_salary_rule.py
```diff
@@ -178,77 +178,84 @@ def _recursive_search_of_rules(self):
...
 
 %s
 """
-                    )
-                    % (self.name, self.code, repr(ex))
                 )
+                % (self.name, self.code, repr(ex))
+            )
+
+    def _get_rule_dict(self, localdict):
```
nimarosa, line 249: Why is the use of this method? Should this logic be inside `_compute_rule_code`? Is there a reason to have this in a diferent method?

nimarosa (changes requested):
I've done a quick review. I'm not sure if the current refactoring in `_compute_payslip_line` affects my private modules. Will test it tomorrow. 

Also i think we have to review the new functions names and organization, something seems odd to me. But will see in depth tomorrow. 

Also interested in @mtelahun review, but i think we can improve this a little more.

mtelahun:
Hi Henrik,

This seems like an interesting PR. I'm a bit busy right now. I'll try 
and get to it this evening.

payroll/models/hr_payslip.py
mtelahun: I agree. _compute_payslip_line is an accurate description of this method. Let's please keep it.

payroll/models/hr_payslip.py
```diff
@@ -523,29 +523,33 @@ def _get_salary_rules(self):
...
-        amount, qty, rate, computed_name = rule._compute_rule(localdict)
-        rule_total = amount * qty * rate / 100.0
+        # compute the rule to get some values for the payslip line
+        values = rule._compute_rule(localdict)
+        key = rule.code + "-" + str(localdict["contract"].id)
+        return self._get_line_dict(
+            rule, localdict, lines_dict, key, values, previous_amount
+        )
+
+    def _get_line_dict(self, rule, localdict, lines_dict, key, values, previous_amount):
```
mtelahun: In keeping with the theme of descriptive method names I think _get_lines_dict() is a better name for this method.

payroll/models/hr_salary_rule.py
```diff
@@ -178,77 +178,84 @@ def _recursive_search_of_rules(self):
...
-                if "result_qty" in localdict:
-                    result_qty = localdict["result_qty"]
-                if "result_rate" in localdict:
-                    result_rate = localdict["result_rate"]
-                return (
-                    float(localdict["result"]),
-                    float(result_qty),
-                    float(result_rate),
-                    result_name,
+        method = "_compute_rule_{}".format(self.amount_select)
```
mtelahun, line 198: :+1: this is an elegant solution to the question of "how to make salary rule computation as flexible as possible as simply as possible"

payroll/models/hr_salary_rule.py
```diff
@@ -257,34 +264,37 @@ def _satisfy_condition(self, localdict):
                  given contract. Return False otherwise.
         """
         self.ensure_one()
+        method = "_satisfy_condition_{}".format(self.condition_select)
+        return api.call_kw(self, method, [self.ids, localdict], {})
```
mtelahun, line 268: :+1: same

mtelahun (approved):
I've included comments in-line, but mostly I think this is a really great PR. You and @nimarosa are really starting to make OCA/Payroll a first-class module. Kudos to both of you :+1:

nimarosa:
> I've included comments in-line, but mostly I think this is a really great PR. You and @nimarosa are really starting to make OCA/Payroll a first-class module. Kudos to both of you :+1: 

Thanks for the comments 👍 that's the idea. If we continue with this changes we can start migration to 15 with a solid and effective module which can compete with others payroll systems.

nimarosa:
@appstogrow please make the changes and we do a fast final review and we can merge.

payroll/models/hr_salary_rule.py
ghost: Yes, I split the method so that one rule may return values for multiple payslip lines.

* Run `safe_eval()` once, in `_compute_rule_code()`
* I will override `_get_rule_dict()` and call `super()` for each line.

ghost:
Done

nimarosa (approved):
It's okay for me now. Please wait for me to adapt my private modules to this PR (I'm doing it today) and I merge, because I have production instances that are affected with this changes.

PR #75, opened 2022-10-01
[14.0] [IMP] payroll: Views improvements and new functionality

nimarosa:
Hello, in this PR i made some views tweaks that i forgotten to do in previous ones, but the change that is interesting here is the new functionality that allows (in a very simple and straightforward way) to add and use python libraries inside salary rules. 

Basically we have a new method called _get_tools_dict() which can be inherited by any custom module and add any library or object to it. Then the library will be accessible inside salary rules by doing: 

result = tools.math

It's a very simple implementation so i think we can merge this fast. I also added the "math" library by default because it's useful for me in my projects and i think it will be useful to use functions like math.ceil or math.floor inside salary rules. 
If you want to add more libraries or objects, please suggest and we can add them in this PR.

With this change, the power of salary rules can be extended practically to any use case. 

Hope you like it! Regards.

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -1,6 +1,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 import logging
+import math
 from datetime import date, datetime, time
 
 import babel
@@ -483,6 +484,11 @@ class HrPayslip(models.Model):
 
         return {}
 
+    def _get_tools_dict(self):
+        # _get_tools_dict() is intended to be inherited by other private modules
+        # to add tools or python libraries available in localdict
+        return {"math": math}  # "math" object is useful for doing calculations
+
     def _get_baselocaldict(self, contracts):
         self.ensure_one()
         worked_days_dict = {
@@ -502,6 +508,9 @@ class HrPayslip(models.Model):
             "categories": BrowsableObject(self.employee_id.id, {}, self.env),
             "rules": BrowsableObject(self.employee_id.id, {}, self.env),
             "result_rules": BrowsableObject(self.employee_id.id, {}, self.env),
+            "tools": BrowsableObject(
+                self.employee_id.id, self._get_tools_dict(), self.env
+            ),
         }
         return localdict
 
--- a/payroll/models/hr_salary_rule_category.py
+++ b/payroll/models/hr_salary_rule_category.py
@@ -19,6 +19,9 @@ class HrSalaryRuleCategory(models.Model):
     children_ids = fields.One2many(
         "hr.salary.rule.category", "parent_id", string="Children"
     )
+    salary_rules_ids = fields.One2many(
+        "hr.salary.rule", "category_id", string="Salary Rule Categories"
+    )
     note = fields.Text(string="Description")
     company_id = fields.Many2one(
         "res.company",
--- a/payroll/views/hr_salary_rule_category_views.xml
+++ b/payroll/views/hr_salary_rule_category_views.xml
@@ -5,14 +5,36 @@
         <field name="model">hr.salary.rule.category</field>
         <field name="arch" type="xml">
             <form string="Salary Categories">
-                <group col="4">
-                    <field name="name" />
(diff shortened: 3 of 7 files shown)


nimarosa:
Please @apps2grow and @mtelahun you are invited to review and make suggestions if you want.

payroll/models/hr_payslip.py
```diff
@@ -483,6 +484,11 @@ def get_current_contract_dict(self, contract, contracts):
 
         return {}
 
+    def _get_tools_dict(self):
+        # _get_tools_dict() is intended to be inherited by other private modules
+        # to add tools or python libraries available in localdict
+        return {"math": math}  # "math" object is useful for doing calculations
```
mtelahun, line 490: I prefer that OCA/payroll returns an empty dict. There is no compelling case to have `math` in the base module but not some other python module. People who need this functionality can override this method in their own modules. If your really need this to be part of OCA/Payroll I would also be okay with creating another module in the repo to provide this functionality separately.

payroll/views/hr_salary_rule_views.xml
```diff
@@ -192,6 +214,108 @@
                             </tree>
                         </field>
                     </page>
+                    <page name="help" string="Help">
+                        <group>
+                            <div style="margin-top: 4px;">
+                                <h3>Salary Rules formula definition (Python)</h3>
+                                <p>
+                                    In python definition in salary rules, you can write any code like you will do
+                                    in python. Any methematical operation or python function is supported.
```
mtelahun: s/methematical/mathematical/

mtelahun (changes requested):
This strikes me as a very powerful tool to make complicated payslip computations, even if I'm not a fan of allowing the implementer even more rope to hang themselves with :grin:

payroll/models/hr_payslip.py
mtelahun: Sorry, I take that back. After thinking about it some more I can see how having some basic methods provided by `math` can be useful without having to install an additional one-line module.

nimarosa: Yes i think math is very useful in salary rules. There are other modules that can be useful but i just keep "math" here because don't want to overload this. Also, it serves as an example of how to inherit and use this method. 

I have some rules that i can refactor using math, that's why i come up with the idea of allowing injecting libraries to the localdict. 

Great you liked it!

payroll/views/hr_salary_rule_views.xml
nimarosa: Fixed the repeated words in last commit, thanks.

nimarosa:
@mtelahun I also added a salary rules views customizations to make them more clean, I think you would like them since you were having render errors with Firefox a time ago.

payroll/models/hr_salary_rule_category.py
```diff
@@ -19,6 +19,9 @@ class HrSalaryRuleCategory(models.Model):
     children_ids = fields.One2many(
         "hr.salary.rule.category", "parent_id", string="Children"
     )
+    salary_rule_ids = fields.One2many(
+        "hr.salary.rule", "category_id", string="Salary Rule Categories"
```
mtelahun, line 23: This should be "Salary Rules", shouldn't it?

payroll/views/hr_salary_rule_category_views.xml
```diff
@@ -5,14 +5,36 @@
...
+                    </h1>
+                    <group>
+                        <group>
+                            <field name="code" />
+                            <field name="parent_id" />
+                        </group>
+                    </group>
+                    <notebook>
+                        <page string="Salary Rules">
+                            <separator string="Asociated Salary Rules" />
```
mtelahun, line 21: Associated

mtelahun (approved):
Thanks. I like the updated look of the views too. No problems on Firefox.

Functional and code review.

payroll/models/hr_salary_rule_category.py
nimarosa: Done

PR #77, opened 2022-10-04
[14.0] [FIX/IMP] payroll: Confirm recomputing and payslip_lines editables

nimarosa:
Hello, this PR brings two new settings, which i unfortunately discovered in production that are necessary... 

1. **Payslip confirm recomputes the payslip.** This is the actual behaviour, but is a very unexpected one since the user will expect to compute the payslip with the "compute sheet" button and then confirm it when it's ready. This automatic recomputation could cause that for some reason the user edits something in the payslip lines manually, and when he confirms the payslip he will lose the changes. I have this unfortunately situation in production with a big quantity of payslips confirmed in bulk, so... you can imagine the usefulness of this. *Now, with this changes, we add a setting for the user to decide whether the confirm button "computes" the payslip again, or it just confirms it.* This setting is by default True (True = Prevent re-compute on confirm) and can be changed if you really need this to be on. For me, it caused a lot of headaches that this was the default functionality. 

2. ***"Payslip lines are manually editable if you see them in the "Computation details" smart button or tree view.*** So by default, the user was allowed to edit manually the payslip lines if he access to the view of the smart button or click in Actions > Computation details. Since the module now ***does not support manual editing of lines*** this was a bug of the view. This also happened me in production when i discovered this tree view was editable. *So now we introduce a new setting that enables/disables the edition of payslip lines.* This is for allowing, in certain controlled situations, to the user to use this editable tree to change something, but in normal cases, lines should not be edited manually so this settings has a default value of "False". 

Since this are two changes that from a user perspective can cause very big problems and headaches, i would like to merge this fast. But will leave it here until someone reviews it because i'm interested in any suggestion you can have about it. 

Please don't hesitate to suggest anything. 
Kind Regards.

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -176,6 +176,16 @@ class HrPayslip(models.Model):
     allow_cancel_payslips = fields.Boolean(
         "Allow Cancel Payslips", compute="_compute_allow_cancel_payslips"
     )
+    prevent_compute_on_confirm = fields.Boolean(
+        "Prevent Compute on Confirm", compute="_compute_prevent_compute_on_confirm"
+    )
+
+    def _compute_prevent_compute_on_confirm(self):
+        self.prevent_compute_on_confirm = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("payroll.prevent_compute_on_confirm")
+        )
 
     def _compute_allow_cancel_payslips(self):
         self.allow_cancel_payslips = (
@@ -224,7 +234,10 @@ class HrPayslip(models.Model):
         return self.write({"state": "draft"})
 
     def action_payslip_done(self):
-        if not self.env.context.get("without_compute_sheet"):
+        if (
+            not self.env.context.get("without_compute_sheet")
+            and not self.prevent_compute_on_confirm
+        ):
             self.compute_sheet()
         return self.write({"state": "done"})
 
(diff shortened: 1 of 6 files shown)


nimarosa:
@apps2grow @mtelahun If you can review it quick in some free time, i would be glad, because i need it in prod but want you to see it first.

nimarosa:
@mtelahun Please also check this one if you want so i can merge today. Thanks for your reviews (Y)

nimarosa:
Merging since not responses and it's a crucial feature in my opinion to prevent users to edit things they don't meant to edit. 

Since the change introduces everything in a setting it's easily customizable and do not generate any problems for the end user. 

/ocabot merge patch

mtelahun:
@nimarosa I appreciate your enthusiasm, but this is not really fair to the rest of us. You put up two PRs at the same time and while we were busy discussing one PR you merged the other one without any review and within less than 24 hours or proposing it. There was nothing in this PR that couldn't wait a few days to be merged. This is a community project. Many of us have day jobs and do not have the resources to be devoted full-time to maintaining OCA modules. I don't think it's too much to ask that you give people time to review changes before you merge them.

nimarosa:
@mtelahun you are totally right. I apologize about it. I really thought that it has been more time since I opened the PR and I had a big issue in prod regarding this so I acted under pressure because the instances affected only consume OCA and I could not change it in private. 

I apologize again to you and the community. 

I can rollback this if someone is really affected by it but I tested it a lot to make sure anything else is affected and since the changes were setting easily changeable i thought it will be right to merge since it has no impact on actual functionality, it only changes view editing and confirm actions. 

Hope you and the others can understand the situation, I really don't had bad intentions about it and thought it will be useful to others to have it fast.

mtelahun:
@nimarosa Thanks. Apology accepted. No need to back it out.Message ID: 
***@***.***>

PR #79, opened 2022-10-04
[14.0] [FIX] payroll: remove total compute method in favour of decimal precision 

nimarosa:
As we always as programmers know, floats are very problematic when we have to deal with precision calculations, also, odoo decimal precision rounding do this worse. 

In this PR I removed decimal precision from the "amount" field in payslip lines. This is motivated by the necessity of precise calculations in payslips. All other fields remains with precision rounding, including "total" computed field which is the field that is displayed in payslips in most of the cases. "Amount" field is used for calculations and in the most of the cases, correspond to the value in "result" variable in salary_rules. Having precision in this field give us unprecise calculations for rules since odoo applies the rounding and we end up with weird values in some cases. Consider this example for more clarification: 

Salary Rule: 
Suposs contract.amount = 104779 and qty = 21 days
```
result = contract.amount / worked_days.WORK100.number_of_days  # This give us the daily unitary amount of a fixed salary
result_qty = worked_Days.WORK100.number_of_days # We use as qty the number of days worked
```

In a case that the employee has worked full time without absences, we should get `total = contract.amount`
The problem is that odoo applies rounding on "amount" field so we end up with the following results: 

```
contract.amount = 104779
unitary_amount = 4989.48 (rounded by odoo decimal precision) 
tota = 4989.48 * 21 = 104779.08 ≠ 104779
```

So as you can see, decimal precision in amount field give us incorrect values even when the calculation is simple (104779 / 21 * 21 SHOULD BE EQUAL TO 104779 and not 104779.08) 

So i think a solution for this is to not round "amount" field which is often a calculated value which we need to continue calculating next salary rules. 

--------------------

The only problem with this change, is that now in the database the values will be stored as they are, giving us values like this: 
Screen Shot 2022-10-04 at 01 46 39

This values are not a problem for calculations since they give us more precision but they are ugly to see. On the other hand, the UI continue reflecting rounded values like here: 
Screen Shot 2022-10-04 at 01 47 32

So from the user point of view, it means no change. The only thing we should take into account is:
If we are going to use this value in reports, maybe we should round it before using it. I couldn't find other problem with storing "amount" this way. If you find any, please suggest changes or another idea on how to mitigate this errors. 

I know that the rounding don't affect the "total" a lot, but in payroll in general, values have to be the more exact as possible since sometimes people receiving the payslips don't understand why the value don't show like it should to be. 

Feel free to make comments, suggestions or ideas about this. 
I keep posted. Regards.

### DIFF
--- a/payroll/data/hr_payroll_data.xml
+++ b/payroll/data/hr_payroll_data.xml
@@ -6,6 +6,14 @@
             <field name="name">Payroll</field>
             <field name="digits">2</field>
         </record>
+        <record
+            forcecreate="True"
+            id="decimal_payroll_amount"
+            model="decimal.precision"
+        >
+            <field name="name">Payroll Amount</field>
+            <field name="digits">4</field>
+        </record>
         <record forcecreate="True" id="decimal_payroll_rate" model="decimal.precision">
             <field name="name">Payroll Rate</field>
             <field name="digits">4</field>
--- a/payroll/models/hr_payslip_line.py
+++ b/payroll/models/hr_payslip_line.py
@@ -32,11 +32,13 @@ class HrPayslipLine(models.Model):
         "hr.contract", string="Contract", required=True, index=True
     )
     rate = fields.Float(string="Rate (%)", digits="Payroll Rate", default=100.0)
-    amount = fields.Float(digits="Payroll")
+    amount = fields.Float(digits="Payroll Amount")
     quantity = fields.Float(digits="Payroll", default=1.0)
     total = fields.Float(
+        compute="_compute_total",
         string="Total",
         digits="Payroll",
+        store=True,
     )
     allow_edit_payslip_lines = fields.Boolean(
         "Allow editing", compute="_compute_allow_edit_payslip_lines"
@@ -68,6 +70,11 @@ class HrPayslipLine(models.Model):
                 )
             else:
                 line.parent_line_id = False
+
+    @api.depends("quantity", "amount", "rate")
+    def _compute_total(self):
+        for line in self:
+            line.total = float(line.quantity) * line.amount * line.rate / 100
 
     @api.model_create_multi
     def create(self, vals_list):
--- a/payroll/models/hr_payslip_line.py
+++ b/payroll/models/hr_payslip_line.py
@@ -35,10 +35,8 @@ class HrPayslipLine(models.Model):
     amount = fields.Float(digits="Payroll")
     quantity = fields.Float(digits="Payroll", default=1.0)
     total = fields.Float(
-        compute="_compute_total",
         string="Total",
         digits="Payroll",
-        store=True,
     )
     allow_edit_payslip_lines = fields.Boolean(
         "Allow editing", compute="_compute_allow_edit_payslip_lines"
@@ -70,11 +68,6 @@ class HrPayslipLine(models.Model):
                 )
             else:
                 line.parent_line_id = False
-
-    @api.depends("quantity", "amount", "rate")
-    def _compute_total(self):
-        for line in self:
-            line.total = float(line.quantity) * line.amount * line.rate / 100
 
     @api.model_create_multi
     def create(self, vals_list):


nimarosa:
The other approach that comes to my mind is changing "total" field from being computed to being a normal field. This way I think (I didn't tried) total will be calculated only on salary rule computation with "real" values of amount field, before it's formatted and rounded by odoo so we should arrive to the correct calculation. This way we can mantain two digits in db fields without losing precision in calculations. 

Please let me know what you think.

mtelahun:
My own view is that the precision is needed to make the calculations deterministic and repeatable. It's more important that one can have confidence at arriving to the same conclusion every time, on every system than that the conclusion is correct :grin: 
I would prefer that you make the precision digits bigger instead of getting rid of it.

nimarosa:
@mtelahun what about removing the computed field in total? I think it has to be more with that than with precision.

mtelahun:
I only glanced at the diff. Let me take a better look at it tonight. See 
if I can come up with a suggestion.

nimarosa:
@mtelahun Actually changed the commit to removing the computed field in "total". This fixes the problem. I think the problem was because rule computation assigns total in _get_payslip_line method but then in the view, the compute method is triggered and is doing the calculation from the rounded values, which leaves us with incorrect values. 

I think removing this compute method does no harm because lines will not be edited manually, and will be always computed by "Compute Sheet" button. Please check if you think this is a better approach. For me it is because we don't mess with precision. Now without this compute method, precision is done after save of the lines, so the calculations on the fly are with the real values.

mtelahun:
After taking a closer look at the problem I realized you have the same 
problem I had. In my country there is no hourly wage. Everyone is paid 
by monthly salary. So, what organizations have done is exactly what you 
did here. They divide the salary by the number of working hours to get 
the concept of an "hourly wage". I said concept here because it has no 
legal standing. The employee's contract specifies only a monthly salary. 
But in that case how do you pay an employee according to the hours they 
worked? The elegant solution would be what you did here: set qty to # of 
hours worked, set amount to hourly wage and let Odoo do the calculation. 
However, this leaves a problem. Many times the number isn't evenly 
divisible into a monetary amount. So when you add the number back up you 
get a different answer, as you said. On the other hand monetary amounts 
only have two decimal places: to take your example you can't have a 
daily wage of $4,989.476190476. I think storing a daily wage like this 
is conceptually wrong. It is also wrong to show the user two decimal 
places, but use the full float in calculations because the numbers 
displayed to the user should add up correctly (i.e. 4989.48 * 21 is NOT 
104779.00).

I got around this problem by instead subtracting the wage for the absent 
hours from the monthly salary. Off course this leaves you with the same 
problem, but in the opposite direction, when subtracting the entire 
month: you get a final result like -0.08. But I easily got around that 
by putting the calculation in

``max(0, $amount)``

I think this is a better solution because:

1. Conceptually (and probably legally) this better reflects the actual 
operation. If an employee is paid by monthly salary it is more correct 
to say that you are subtracting a monetary amount for absent hours.

2. The numbers on the payslip are mathematically accurate.

3. When it's done this way the employee understands and is more 
accepting of why you have to round the daily wage to two decimal places 
and can't use the more accurate float amount in the calculation.


Regards,

Mike.

nimarosa:
@mtelahun Hello Mike, yes we had the same problem. And in most of the cases, i've done like you. In reality, in my country you show one line with "Monthly Salary" showing the full number, and then we use another salary rule to subtract the amount in another line, so the employee can see his full salary and the subtracted amount just in the second line. So this is not a problem now. (Also this is the reason of negative values in worked days ;) )

The problem comes with, according legal standards of payslips in my country, we have to show the number of days of the month in the quantity amount, so even if we don't subtract anything from the full salary, we still need to divide it by the number of days worked and then multiply it using result_qty. This way, we end up with the same number but we show in result_qty the correct amount of days. If we don't do it that way, we always going to show "1" in quantity which is wrong and not valid. 

Anyways, i discovered the error it's not in the calculation of the salary rules, it is in the computed field for "total". I think this was a computed field because the module allowed in the past to edit manually the payslip lines, so this field was computed to facilitate the calculations. Since nowadays we don't let the user edit the payslip lines, i think we can remove this compute method. 
Removing it, let salary rules calculate the amounts, and if you see the code, the salary rules do not use rounded values. The rounding happens when the rules are created. So if we have this computed field, it's this calculation that shows the wrong value. Look at last commit, if we remove this computed method from total, we have precise calculations, even in the case you are describing i think this will solve your use case too.

mtelahun:
@nimarosa What about the problem of the payslip lines, as displayed, not being mathematically accurate? That's a deal-breaker for me.

nimarosa:
@mtelahun they are mathematically acurate. Because we restore the decimal precision to all fields. They will be more acurate than before because the compute method was re-computing total again on form view using db values so that was the issue that causes the inaccuracy on the first place. 

Please try this PR and you will see that your numbers will be the same or even more accurate than before because now the system don't re-compute total two times. 

Before this total was calculated in salary rules and also recalculated by the computed field which causes that. 

Also categories sum values will be more accurate that was another issue of that this compute field. Sum of categories were a bit different to sum of displayed values. 

I made several tests and everything is computing okay with decimal precision in all fields and more accurate than before.

mtelahun:
@nimarosa By mathematically accurate, I meant that if I took out a calculator and manually did the computations shown on the payslip lines I will get the same answer as the software. If I don't, then the software has committed one of the cardinal sins by breaking the first commandment of payroll processing: The employee, performing the calculations manually, using the numbers shown on his payslip, shall arrive at the **same exact** answer as the software.

This is with your patch:
Screenshot from 2022-10-04 18-43-00

mtelahun:
On 04/10/2022 16:52, Nicolas Rodriguez Sande wrote:
>
> The problem comes with, according legal standards of payslips in my 
> country, we have to show the number of days of the month in the 
> quantity amount, so even if we don't subtract anything from the full 
> salary, we still need to divide it by the number of days worked and 
> then multiply it using result_qty. This way, we end up with the same 
> number but we show in result_qty the correct amount of days. If we 
> don't do it that way, we always going to show "1" in quantity which is 
> wrong and not valid.
>
Circling back to this.

I'm not trying to argue about payroll rules in your own country, but for 
me if the employee is being paid a monthly salary and payroll is paid 
every month (as opposed to by-weekly) then yes, 1, is the correct 
quantity. If the employee is an hourly employee then the quantity should 
be # of hours, and if he's paid on a daily wage the quantity should be # 
of days.

nimarosa:
@mtelahun yes, the logic dictates that must be one. But my country is a very special one and here we must show total days in quantity because a lot of calculations are based on daily rate or hourly rate depending on the case, but sometimes the labor unions gives the salary in monthly basis so you have to proportionate the amount. 

I don't know how your payslip reports are but in mine the only value that you shown on payslip is quantity, rate and total. So the employee can do the calculations because you only use amount to internal calculations.

Anyways I don't argue about that you need this in your country. Maybe you will be comfortable with a setting that for that? So you can choose to use it or not? Sorry to insist but I have clients very concerned about this because it happens the contrary, the employee don't understand why the total is not equal to the actual salary, so we have a problem in both ends haha. I agree with you for me it should be like you said, but it's true that this problem affects the employee understanding of the payslip, at least for my cases of use. 

Will you agree to create a setting?

nimarosa:
Also I noted that why is the usage of salary rules calculating the total (see get payslip lines function) if then is going to be recomputed by the computed field? In that case we are computing twice which I think it's no right.

mtelahun:
@nimarosa I can't support this, even with a setting, because it completely bypasses the payroll decimal accuracy and you can already achieve what you want with the current payroll module by simply increasing the Payroll decimal accuracy. This is what you get when you increase the Payroll decimal accuracy to 4:
Screenshot from 2022-10-04 20-54-27

mtelahun:
> Also I noted that why is the usage of salary rules calculating the total (see get payslip lines function) if then is going to be recomputed by the computed field? In that case we are computing twice which I think it's no right.

It needs to be calculated in the rules so that you can reference the total from previous rules in subsequent ones. It needs to computed again by the computed function so that the calculation can be done respecting the decimal accuracy. In fact, I would argue that the calculation being done in the rules has a bug because it is not respecting the decimal accuracy setting. But trying to fix that at this late date will cause rules that may have been working for years to suddenly start giving slightly different answers so probably not a good idea.

nimarosa:
> It needs to be calculated in the rules so that you can reference the total from previous rules in subsequent ones. It needs to computed again by the computed function so that the calculation can be done respecting the decimal accuracy. In fact, I would argue that the calculation being done in the rules has a bug because it is not respecting the decimal accuracy setting. But trying to fix that at this late date will cause rules that may have been working for years to suddenly start giving slightly different answers so probably not a good idea.

@mtelahun yes that's what I noted, you are calculating using different values inside salary rules and then in the actual total column the calculation has a different result. 

I agree with you that we need the decimal accuracy but in your example even with "4" accuracy setting you continue having decimals in total column. 

Also using some Max or math function for this does not solve the problem because you don't know if the result will be less or more of the actual expected value. So you can't have a rule which generally fixes this inside the rule. 

Maybe we can have a separate decimal accuracy for amount field and mantain the "2" accuracy for the total and quantity field... will continue trying to solve this in a way we don't change accuracy but receive the correct values. 

Will get back to you with new changes.

nimarosa:
@mtelahun Created a separated decimal precision object and applied it to "Amount" field. Is this acceptable for you? Having "amount" with more accrual precision algo solves the problem.

mtelahun:
Two decimal places works for most people and most use-cases. This is 
something that properly belongs in a private repo, not in OCA.

nimarosa:
Closed for now. If I find a better way to improve calculations for accuracy will open another. For now I resolve it on my end.

PR #80, opened 2022-10-13
[14.0] [ADD] payroll_hr_public_holidays module 

nimarosa:
Hello friends, 

This time i bring a new module. It's a simple module that links hr_holidays_public (from OCA/hr-holidays repo) with payroll. Basically this creates a line for public holidays in worked days table in payslip form. 

This is useful because in some countries we need to have specific salary rules to pay public holidays or make complex calculations. As the public holidays are automatically deducted from WORK100 entry, it might be useful to have a linea with number of days and hours of public holidays. 

Hope you find it useful. 
Regards.

### DIFF
--- a/payroll_hr_public_holidays/models/hr_payslip.py
+++ b/payroll_hr_public_holidays/models/hr_payslip.py
@@ -1,7 +1,6 @@
 # Copyright (C) 2021 Nimarosa (Nicolas Rodriguez) (<nicolasrsande@gmail.com>).
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from datetime import datetime, time
 
 from odoo import _, api, models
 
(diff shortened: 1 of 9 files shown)


payroll_hr_public_holidays/models/hr_payslip.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @api.model
+    def get_worked_day_lines(self, contracts, date_from, date_to):
+        res = super().get_worked_day_lines(contracts, date_from, date_to)
+        for contract in contracts.filtered(
+            lambda contract: contract.resource_calendar_id
+        ):
+            day_from = datetime.combine(date_from, time.min)
+            day_to = datetime.combine(date_to, time.max)
+            day_contract_start = datetime.combine(contract.date_start, time.min)
```
mtelahun: I don't understand why this is a datetime. I don't see anywhere where a datetime is needed. It would be cleaner to just use a date type.

payroll_hr_public_holidays/models/hr_payslip.py
```diff
@@ -0,0 +1,47 @@
...
+    def _compute_public_holidays_days(self, contract, day_from, day_to):
+        # get public holidays list
+        public_holidays = self.env["hr.holidays.public"].get_holidays_list(
+            year=day_from.year,
+            start_dt=day_from,
+            end_dt=day_to,
+            employee_id=contract.employee_id.id,
+        )
+        ph_days = len(public_holidays)
+        ph_hours = ph_days * contract.employee_id.resource_calendar_id.hours_per_day
```
mtelahun: What do you think about providing an alternate default value for hours per day in case the employee doesn't have a resource calendar?

mtelahun (changes requested):

payroll_hr_public_holidays/models/hr_payslip.py
nimarosa: It's a datetime because this fields come from payslip. Look at original function I just replicated what is in hr_payslip to have the exact same values that other worked days functions use. 

I think it's a datetime because I allows to use relativedeltas and that's why payroll module use it. It's makes calculation better. In this case, it's not necessary to use datetimes but its payroll module who is using it.

payroll_hr_public_holidays/models/hr_payslip.py
nimarosa: Yes I thought about that. But then I thought that if employee don't have a resource calendar the module will not compute leaves and worked days. So we can just don't return anything if employee has no resource calendar or make it use a fixed value like 8 hours (it can also be a setting if you think it's useful). But I think payroll module requieres a resource calendar to compute worked days so in the 99% of the cases the employee should have one.

payroll_hr_public_holidays/models/hr_payslip.py
nimarosa: Done.

payroll_hr_public_holidays/models/hr_payslip.py
nimarosa: @mtelahun I adoped 8 as default value since it's the more common work schedule. Anyways if employee has calendar, it will use the correct one.

mtelahun (approved):
Looks Good: code review

Issue #78, opened 2022-10-04
About pending migrations


nimarosa:
Hello oca/payroll users and contributors. 

This thread is to discuss migrations to v15 and v16. As you can see, the new v16 branch is now live and we still in the v14.

Since we are using v14.0 to do improvements and changes to the modules, I suggest we continue doing so until we think it's time to migrate to v15.0 and then to v16.0. 
My approach here will be maintain the tree versions because i have clients that will continue with v14.0 for a while, but i think we should start with v15.0 migration soon. 

From my end, i have some PRs that i wish to merge first in v14.0 before we start the migration so we can have this changes in following versions. I will try to include all this changes this week or the following.
If someone has something new or improvements that wish to include in the module, i suggest to do it as fast as possible.
So i think we can schedule deadline for starting migrations to other versions on 20/10/2022. 

As i said, i will continue maintaining v14.0 version but also after migration, I will port any change in any branch to the other ones, so it will be good for all of us to include any changes you wish to include before this date, so we reduce the workload of backporting versions. 

Any doubt or suggestion about migrations can also be addressed in this thread. 
Kind regards to everyone.

TheCloneMaster:
Hi @nimarosa,
We have made a couple changes for multicompany payroll management, to allow payroll admins to create runs for employees on different companies without switching their own. Company_id is calculated from employee's own company, not from env's company default.
Also to allow analytic account from employees' contract if defined instead of rule's account.
Both are extra features, if not configured behavior doesn't change.
Do you consider them useful, to present PRs?

Regards,
-Mario

nimarosa:
> Hi @nimarosa,
> 
> We have made a couple changes for multicompany payroll management, to allow payroll admins to create runs for employees on different companies without switching their own. Company_id is calculated from employee's own company, not from env's company default.
> 
> Also to allow analytic account from employees' contract if defined instead of rule's account.
> 
> Both are extra features, if not configured behavior doesn't change.
> 
> Do you consider them useful, to present PRs?
> 
> 
> 
> Regards,
> 
> -Mario

Hello @TheCloneMaster, yes feel free to introduce the PRs and we can review it. It looks a useful functions for businesses that use multicompany.

mtelahun:
Odoo is now at 16.0 and we haven't yet migrated modules to 15.0. Is anyone working on this? I have an interest in moving to 16.0 soon so if no one is working on this I will propose PRs for both 15.0 and 16.0.

pedrobaeza:
[...]

PR #81, opened 2022-10-13
[15.0][MIG] payroll

mtelahun:
Fixes: #19 
cc: #78


mtelahun:
It looks like there's a build problem. Probably a stale config somewhere. Can someone take a look please?
@nimarosa @pedrobaeza

mtelahun:
I mean the build failures. I'll take care of the pre-commit failures.

mtelahun:
@nimarosa @pedrobaeza Sorry. It was my own fault for not checking the migration went properly.

nimarosa:
@mtelahun i think contract views changed a bit in v15.0 that's why build error is about a xpath not found.

nimarosa:
@mtelahun Please before finishing this do a rebase with 15.0 version since i pushed updated dotfiles to have v15.0 up to date.

norlinhenrik:
Can we replace `hr.rule.parameter` & `hr.rule.parameter.value` with a dependency on [base_time_parameter](https://github.com/OCA/server-tools/pull/2426)?

nimarosa:
@norlinhenrik Hello Henrik, I think adding a dependency to a server-tools module will be bad for payroll. 

Most people don't download all oca repos in his instances so it will end up with an dependency error when trying to install payroll and not having the other module. I think it's good to maintain the less dependencies required here. 

Also, and this one is more personal, the way I see it the user who uses rule parameters expect to have the view inside payroll module. Doing this with an inherited module will be a little tricky because we need to filter all views by domain and add again the module field in base_time_parameter and think of a way to integrate it in payroll the way it's done now. 

In my instances I use time parameter to store specific payroll data that has to be easily changed by the user and also have some custom implementations around this model and no time to migrate them now to a new behavior or module. 

If you have a good way of doing this by leaving the current functionality intact you can propose a PR on the 14.0 branch and it will be migrated to 15. 

I'm not a fan of introducing big changes in migrations because it causes confusion and errors and will requiere migration scripts, that's why I wanted to do the big changes in 14.0 so we make modules maintenance more easy and straightforward. 

Hope you understand and feel free to suggest a PR if you can do it without affecting current functionality.

mtelahun:
@nimarosa All is green now. Sorry for the noise. This was the first migration I did in a *long* time.

During my migration I noticed that Odoo has removed the `code` field from hr.leave.type. So, I have reverted the payroll module to using the `name` instead.
Maybe, starting from 15.0 onward we need a new model to map hr.leave.type to a code in worked days. What do you think?

mtelahun:
> Can we replace `hr.rule.parameter` & `hr.rule.parameter.value` with a dependency on [base_time_parameter](https://github.com/OCA/server-tools/pull/2426)?

I think this is a separate issue and belongs in a different PR

nimarosa:
> During my migration I noticed that Odoo has removed the `code` field from hr.leave.type. So, I have reverted the payroll module to using the `name` instead.
> 
> Maybe, starting from 15.0 onward we need a new model to map hr.leave.type to a code in worked days. What do you think?

Hello @mtelahun. I hate when odoo changes things like this, but I think is has to be because enterprise payroll module didn't use that code because enterprise module is different (our module is better un my opinion). 

What do you think about inheriting hr.leave.type here in the payroll module and adding this field again? That way we can maintain current functionality that is key for the module, at least for me. I think using name in salary rules will generate problems. 

If you think that adding this field to leave type will fix it we can include it in this PR.

nimarosa:
@mtelahun also noticed you removed all string="" of a lot of fields. Why is that?

mtelahun:
Yes, that's a better solution. I'll do that.

mtelahun:
pre-commit errors:

- string attribute on `<tree>` view is no longer allowed

- in model creation string attribute on a field that is the same as the 
default string provided by the ORM are no longer allowed

nimarosa:
@mtelahun yes but you are removing string in model creation not on tree view. Correct me maybe I didn't understand

mtelahun:
Please re-read my reply. There are two places where the `string=` 
attribute raises errors.

1. in `<tree>` view

2. in model creation when the `string=` attribute is the same as the 
default one provided by the ORM

nimarosa:
@mtelahun great I understand now. 
Please tell me when it's ready so I review and merge.

Also if you can review the fix I made in 14.0 we can merge that first and include this change in following versions too.

mtelahun:
@nimarosa it's ready for review.
No problem, it can wait for #83

payroll/models/hr_payslip.py
```diff
@@ -0,0 +1,803 @@
...
+            day_from, day_to, calendar=contract.resource_calendar_id
+        )
+        for day, hours, leave in day_leave_intervals:
+            holiday = leave[:1].holiday_id
+            current_leave_struct = leaves.setdefault(
+                holiday.holiday_status_id,
+                {
+                    "name": holiday.holiday_status_id.name or _("Global Leaves"),
+                    "sequence": 5,
+                    "code": holiday.holiday_status_id.name or "GLOBAL",
```
nimarosa: @mtelahun Okay so here you continue using name. Also i think you have to add the new file adding the field to hr.leave.type.

I would prefer we include this change in this PR because its a crucial functionality of the module.

nimarosa (changes requested):

payroll/models/hr_payslip.py
mtelahun: Done

mtelahun:
@nimarosa @apps2grow Please review. It's ready now and includes recent changes in 14.0

nimarosa:
@mtelahun Hello, we will wait to close/merge #87 that has to be changed to create a new module and remove the parameters functionality from payroll. Then we can rebase and we start merging the migrations.

mtelahun:
@nimarosa @apps2grow Please review and let's merge. This incorporates the latest changes.

nimarosa:
@mtelahun for me it's okay. Just done a code review and commit history looks okay. 

Did you tested it yet? Like functional.

nimarosa (approved):

norlinhenrik (approved):
I successfully computed a payslip for Marc Demo.

nimarosa:
Okay let's go. 

I have a few PRs that I might add to 14.0 but we can migrate them later.

/ocabot merge nobump

mtelahun:
@nimarosa Yes, I tested it. Thanks for merging.


Mike.

PR #87, opened 2022-10-17
[14.0] [ADD] payroll_rule_time_parameter / [IMP] payroll: remove rule_parameters from payroll & [FIX] payslip and payslips objects

norlinhenrik:
With this PR, payroll depends on [base_time_parameter](https://github.com/OCA/server-tools/pull/2427).
@nimarosa if you don't want to depend on server-tools, would you like to copy base_time_parameter into a private repo?

What do you think about these **changes in models and fields**?
* hr.rule.parameter --> base.time.parameter
  * parameter_version_ids --> value_ids
  * new field: type
* hr.rule.parameter.value --> base.time.parameter.value
  * rule_parameter_id --> parameter_id
  * parameter_value --> value_text
  * new field: value_reference
  * new field: type

**Would you prefer** `base.time.parameter.version` and `version_ids`?

Later I will write migration script for this PR.


nimarosa:
Hello @norlinhenrik, if you want to make this working i think you should implement the part of hr_salary_rule_parameter as implemented previously, (see hr_payslip.py file) it adds the function to Payslips object to allow calling it from inside salary rules using payslip.rule_parameter("CODE").
 
Also the views of the salary rules parameters should be by default filtered to only "Payroll" records type of parameter, so if we use the module in another place, we don't end up with all time parameters of the instance. For this i think we should add a "Model" field that allows us filtering the views. 

We can work it out in this PR but we need to replicate current functionality to change hr_salary_rule_parameter to this new dependency.

nimarosa:
Also check the runboat and tests since they are not passing.

norlinhenrik:
@nimarosa Now you can test this PR together with the PR for [base_time_parameter](https://github.com/OCA/server-tools/pull/2427).

nimarosa:
@norlinhenrik Can you rebase this so i can test it?

nimarosa:
@norlinhenrik We have a licenses issue between modules. I suggest we choose LGPL-3 license for both modules this should fix tests. 

`payroll (LGPL-3) depends on base_time_parameter (AGPL-3)
`

pedrobaeza:
Both should be LGPL instead. As payroll comes from Odoo, we can't relicense it.

nimarosa:
@pedrobaeza @norlinhenrik so we have to make PR changing the license in server tools module and then we retrigger tests here.

pedrobaeza:
Yes

nimarosa:
@norlinhenrik Just retriggered the tests. You also have a error in pre-commit. 

`payroll (Beta) depends on base_time_parameter (Alpha)` I think we will have to change alpha status of base_time_parameter

norlinhenrik:
@nimarosa I was thinking to make a test to satisfy covecov. But then I have a question:

Now when hr.rule.parameter is renamed to base.time.parameter, would you also like to rename payslip.**rule**_parameter(code) to payslip.**time**_parameter(code)?

nimarosa:
@norlinhenrik i would prefer leaving the name as rule_parameter since I have a lot of rules that use this method and people are trained to call this method. Changing it now it will have impact in current instances. 

What we can do is add a deprecation warning log message and we can change it maybe in the future. 

I will try to make time to test this today, so we can merge. A question, is the views filtered by model solved?

norlinhenrik:
@nimarosa I have followed your suggestions now.

payroll/views/base_time_parameter_views.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="base_time_parameter_action" model="ir.actions.act_window">
+        <field name="name">Time Parameters</field>
```
nimarosa: @norlinhenrik  Last suggestion. Can we mantain the name of the view as "Salary rule time parameter" or something similar? 

It's important to me that my current users of the module adapt to this change with the minor hassle possible. Apart of that, I like the change, will test it when the other module PR is merged and we can merge here.

nimarosa (changes requested):

payroll/views/base_time_parameter_views.xml
norlinhenrik: Done

norlinhenrik:
@nimarosa Please test the migration.

* Use branch 14.0 to create a database with a parameter.
* Change to branch 14.0-pr87 and manually set the version to "14.0.6.0.0".
* Update payroll and check that the parameter is correctly migrated.

I tested this already, but I recommend that you do it also.

mtelahun:
@apps2grow 
Hi Henrik, is there any specific reason why this should go in payroll? This seems like an optional feature that belongs in another module, especially since it makes payroll dependent on a module in another repo. Why should people who have no reason to use this feature now be required to install another module?

nimarosa:
@mtelahun Hi, this functionality is already in payroll. It's the salary rule parameters functionality introduced by me a time ago. It's used to create parameters that can be accessed inside rules and have values that change depending of the date. 

In Latin American countries is widely used since our local taxes, amounts and salary scales are updated more than 3 times in the year. With this functionality you can have parameters which have versions with a date. Depending on the date of the parameter and payslip, is the value returned. 

With Henrik we noted this was an useful model for other use cases (he uses it in the tax module). So we created base_time_parameter in server-tools repo. Since base_time_parameter give us the same functionality that is replicated now inside payroll in files salary_rule_parameters and salary_rule_parameters_versions it makes sense to use the module as dependency since it's more advanced than the current implementation I've made here in payroll and so remove replace it with this dependency. 

But it's no anything new, this functionality is already in payroll without the dependency. 

Also installing the base_time_parameter makes no harm to the use who don't want to use this functionality. It's a server-tools module and have no impact in anything if you don't use it. 

I don't like to add dependencies without a reason too, but I think due to the advance in the module and the work of Henrik to integrate in payroll that make sense to do this change. If we have a module that does the functionality we need, why should we copy paste here in payroll instead of use the module as a dependency? I think it makes sense since it's a great functionality at least for the uses cases I have seen, and using this dependency will also make easy to maintain payroll not having to copy-paste code that is in our other module.

mtelahun:
Yes, I understand that. And at the time I wasn't really happy about it 
going into the payroll module, but I chose not to say anything because I 
only have so much time to dedicate to OCA/Payroll and didn't want to 
burn cycles about a feature I had no plan to use. But now you are 
talking about introducing a dependency that *everyone* who uses payroll 
will have to install even if they are not using this functionality. The 
payroll module is *the* central module of the Payroll repository and 
users shouldn't be forced to install a third module from *another* 
repository just because it's needed by an optional feature that they 
don't use. The proper place for an optional feature that introduces 
dependencies on other modules is in a separate module.

nimarosa:
@mtelahun Yes you are right but with this logic there are more functions that some people don't use (not only here but in a lot of odoo modules) like contribution registers, salary rule inputs, child rules, and that's not a reason to remove them from the module. I think it makes no harm to the module to have functionalities that can be optionally used. 

Also we don't know exactly how many people use it or not, that's why a migration script is included. We only know our own use cases but not no one knows the whole metric. 

Payroll module has a lot of downloads since we started making changes, that tells me that we are improving it in the right direction. If we make a new module for every new functionality we will end up with a big repo and make more complex the installation of the payroll suite. 

Also when I introduced this new functionality the majority of people including Pedro agreed to put it inside payroll. So we have to do what the majority of maintainers and users want. 

Also, almost all server-tools modules are meant to be used like dependencies, that's the whole point of the repo, use this modules to plug-in generic functionalities to other modules. Odoo dependency system for me it's good, and anyone who use OCA often clone server-tools repository so they will install the module with no problems when installing payroll. 

I think that if we move this to another module now, we will affect users which are using the functionality since they will not know that they have to manually install another module and at least for me parameters are used in several salary rules so changing it will generate production instances to fail. 

My approach with payroll repository is to make a suite of payroll apps and modules that can compete with professional payroll systems and I think we are achieving it, we have more downloads every day and having a complete payroll system in odoo is a very big feature that can induce people to use odoo what I think is what everyone in OCA wants. 

I think that the case that you or some users don't use it don't justify removing it. Like I said, people want a complete a robust module, even if they don't use the functionality it's available for them if they need it and makes no harm to have it there. 

I will vote for having this feature in the module, but like I always say, if you want to suggest another way of achieving this that fit your users needs, you can propose it. 

Like I said I don't like the dependency thing but i understand that if the functionality is achieved by another module it makes no sense to maintain a copy of code here for the only reason of not having a dependency. 

I don't know if it's possible to make a setting that install dependencies when enables, that can be a solution, but never done it. 

I know we can differ in point of views but wanted to take time to make my point to you since you are an active contributor to the repository and I want to listen all point of views, but in this case I really think it makes no harm adding a dependency. The final user will almost not notice the change, since the common practice is to clone server-tools in a new odoo instance because other modules also depend of modules there, so for the user the installation will be like always.

payroll/views/base_time_parameter_views.xml
nimarosa: @norlinhenrik Thanks Henrik, will test everything today and let you know.

mtelahun:
@nimarosa my problem isn't with the feature. My problem is that I and 
others who don't want to use this feature are being forced to install 
another module which we will not be using just so we can run our 
payroll. This is not functionality that is essential to running payroll. 
As far as I am aware moving the code to a new module won't affect the 
migration and any users using this feature will pick up the dependency 
and migration automatically. It is unfair to force others to go through 
pointless hassles that don't affect them just to reduce the work for 
yourself.

nimarosa:
@mtelahun It's not for myself. What i meant is that if other users (not mine) are using the functionality and we change it to a new module, they will lose all current configured parameters since they will have to install a new module to have the functionality again. That process, implies losing the hr_salary_rule_parameters that they have already in their systems. 

As far as i know i'm not the only one using this functionality, since @norlinhenrik is using it as well. And they could be other users that we don't know of because they don't write here. 

If we change this in payroll module, the migration script will take care of this cases, it's not a problem of myself only. If we move everything to another module, which i don't have a problem with, all current users that are using this will lose their data because we can't write a migration from module to module. This in production environments could be a very big hassle. At least as i see it, like odoo works if you remove a model then all tables related to it will be automatically dropped, so even if they install the new module after that, they will lose their data and have to create all parameters again. 

Finally, i understand that you don't consider it essential, but some people might consider it essential. It depends of the use case and you can't affirm that if you don't use it noone use it. I see that you don't find it useful, but you can't know if other users of the module differ. Also i don't think that being "forced" to install a dependency like base_time_parameter does any harm to the end user. The module don't do nothing if you don't use it, it has no cron jobs, or any that can impact performance, so i really don't see the problem. A lot of account modules has server-tools dependencies and it don't seems to be a problem to anyone. 

Anyways i don't want to cycle through a discussion that we will don't have the same point of view. I will test this today and then we can make a vote and tag other people so we can have more voters than the tree of ours. 

@pedrobaeza @norlinhenrik When you have time, maybe you can join the discussion so we can all vote and make a decision about it. The options are: 

- (1) Moving this to a new module. This implies current users losing his information and having to create parameters again. 
- (2) Merging this PR, accept the dependency on the server-tools module. User will don't notice because of the migration script. 
- (3) Leave the payroll module with the current implementation, and maybe in future versions, create a new module. This will leave us with having to maintain duplicated code, more work in migrations and not having the benefit of base_time_parameter new updates and improvements. 

Regards.

nimarosa (approved):

mtelahun:
1. I thought the point of migration scripts is to handle exactly these 
kinds of problems. Why is it not possible to handle this case with a 
migration script?

2. Previously, when I objected to PRs changing the behavior of running 
instances, you seemed to object on the grounds that it was OK to do this 
in 14.0 to allow rapid progress of development. Why is it suddenly so 
important not to inconvenience users when it's your pet functionality we 
are talking about? Functionality introduced only a few months ago, in fact.

3. Your comparison with OCA/account-* is misleading. Those are all 
modules that provide *additional* functionality to base account module. 
In our case there is no base payroll module in Odoo. So, the payroll 
module in OCA/Payroll has to function as the base payroll module for the 
community.

nimarosa:
@mtelahun 

1. Is it possible to do a migration script that migrates data between modules? Because from my point of view, when the user updates payroll, it will drop the parameter database tables so even if they install the new module after, the data will not be migrated because it will be already deleted. That's why I think is not possible. If you think it is, please explain a possible migration script between modules without dependencies. 

2. When you objected that, we were talking about module functionality. Any current data was NOT affected by any of the changes we have been making. If any data was affected I think we will introduce a migration script like this PR is doing. My point at that time was that we can change functionality without affecting any running instances data, which is the important part. 
Also in some PRs you objected something about changing the signature of functions and I accepted your suggestion and refactor everything to not affect current signature of functions because you stated that private modules could be inheriting this functions and expected returned value not to change. I accepted that suggestion, because it was right and even when OCA has no obligation in providing support for external private modules. That change was affecting one of your modules I think and I had no problem in accepting your suggestion. We are here to make the module better not to mess with current users, so everything we can do to change the module without messing with data or current implementations we will do. 

With that said, can you explain why is so bad for you to accept a OCA dependency that will not affect anything if you don't use the functionality? I don't want to fight here but it seems that you don't want any change that you don't want to use. But that's not the case for all users, apart from my instances, I know people who is using the module and the feature. Also I know that when we started making changes to the module we have more downloads and more people using it, even if the didn't post here. So it's my duty as a PSC try to do what is best for the repo. 

3. Like you said. There is no base payroll module in odoo. And this is the open source option to it. So if we are developing a base module, why we shouldn't try to make it the most complete and robust one in odoo community? Why is so bad to have functionalities even if some people don't use them? 

If you see any odoo base module they always have functionalities that people don't use. People use field service and maybe don't use all of its fields and functions. People use CRM and maybe don't use all the prospecting functionalities and use it just to track sales communications.

There are a bunch of use cases for every module, some people might use all functionalities, some might not. But what is the problem of having more functionalities? I think It don't affect the final user to have a dependency that don't do anything if you don't use it. It don't make any security issues, and don't affect performance in any way. So what's the problem? That's is what I don't quite get about what your are saying. 

You said that you have no problem with the functionality itself. Okay. It's the dependency? So according yo you we have to copy all the base_time_parameter module code into payroll and maintain it doing copy and paste when the other module gets improved? I think that is not quite practical. I have no problem doing it but as one of the main maintainers of the module, I think trying to reduce maintenance work even if more maintainers join us, it's not a bad idea. Why should we reinvent the wheel if we already have a module for this functionality? 

---

Finally I want to state that I have no problem if this is a separate module or it's in payroll. I can adapt my modules and clients (but it will delay the merge). The problem is that I'm thinking about people who might be using this functionality and doesn't follow repo PR or issues, people that are not my clients, but I know they use it. Yes I can warn them about it, but... so you really think that all users of payroll are me, Henrik and you? What I'm proposing is to be sure we don't affect users, I don't know how many they are or how many people is consuming the module. The only metric I can see is the increase in downloads that gives me a hint that people started to use payroll (which was orphaned and not maintained for a long time) after the new features and improvements we introduced, so from that I can deduce that it should be people using it.  

In resume, and like I said. I have no problem with any of the solution proposed. Meaning we can make a new module, maintain current implementation or accept this PR. But I think we have to value Henrik @norlinhenrik work here and consider this approach. For me, it's the same, I can adapt myself to any decision, but i really think we should vote and move on with the majority because we have different point of view and I think all of them are valid, so we should appeal to democracy. That's why I invite anyone reading this to vote, but I also will value more the opinion of people who is involved in the repo, that people are you, me, @norlinhenrik and @pedrobaeza so far. 

I really value your opinion Mike, and your suggestions were really useful in past PRs, that's why I'm taking time to write such long responses. But I think we should decide what's the majority of us prefer. I know you have a point, I have another. So let the majority decide but i really think that adding a dependency it's not such nightmare like you are describing, and more if it's a dependency that is maintained by Henrik and me, so we are highly involved in payroll and will not leave that other module without maintenance or do something that could break payroll.

Thank for yours responses, like I said I really value your input but I think we might not come to an understanding here since you and I have two different point of views, both valid, and I think the only way is let the others decide by majority.

mtelahun:
Message ID: ***@***.***> The payroll 
module is the minimum module needed to run payroll in CE; therefore, it 
should be considered a base module. It should only depend on base Odoo 
modules. Depending on a module in another repository means that we lose 
some measure of control over our base module. If we want to make changes 
in our module that require changes in the dependency module we need to 
coordinate with and get agreement of the maintainers of the other 
repository. Any changes in the dependency module will force us to make 
changes in the payroll module even though the original change may not be 
related to payroll. This means we will have to monitor the module in the 
other repository to keep abreast of any changes to make sure it doesn't 
affect us. During every new Odoo version we will have to get the other 
repo maintainers to upgrade the module before we can upgrade our payroll 
module. If no one else is interested in the other module we will be 
forced to keep maintaining a module in another repository for the 
foreseeable future. Right now it may seem like everything is fine and we 
can get the cooperation of the other repo's maintainers. But we can't 
depend on that. What happens when the other repo maintainers change? 
What happens when, in the future, there is a fundamental disagreement 
between us and the other repo maintainers. They don't depend on us so we 
will be the ones forced to make a change we disagree with. In an 
optional module this is fine and part of maintaining an optional 
feature. But for a base module it is unacceptable.

nimarosa:
@mtelahun the other repo maintainers are me and @norlinhenrik, so we are the same people that maintain payroll.

I have no problem with moving it to another module, my concern with that is how we make a migration between modules that don't depend of each other. I also don't have problem in continuing the current implementation having the feature directly in the module. 

Like I said, I understand your point. I think you understand mine. Let's let other people involved here to decide which approach we use.

pedrobaeza:
Can you please summarize the thing "to vote"? The thread is very long.

nimarosa:
@pedrobaeza yes, 

(1) We accept this PR and base_time_parameter (which me and Henrik are maintainers) becomes a dependency of payroll and now the parameters functionality uses this module. This is what @mtelahun don't want. Me and Henrik are okay with it. 

(2) We move the functionality of payroll parameters to another new module here in payroll, where base_time_parameter is a dependency but in a separate module. - For me this is not practical since we can't not make migration scripts between modules so people using the current implementation of parameters will loose the data and even if they install the new module after, will need to re-create the parameters. Also we don't know how many people could be using it so it can cause problems in production. 

(3) We mantain the current functionality as it is now (having base_time_parameter implementation inside payroll) - This stand for more mantainance efforts since we will be coping and pasting code that it's already available in the server-tools module. In other words, reinventing the wheel instead of using a module that provide us the functionality. 

That's a resume so far. If you need more info you can read past responses and the reasons of each part.

mtelahun:
I still don't understand why you think option 2 (move the functionality to another module and use a migration script to migrate data won't work. This is exactly the sort of thing it's supposed to do: copy the data to temporary tables, upgrade the modules, move the data from the temporary tables to the correct place.

pedrobaeza:
I'm also with option 2, as we don't "bloat" the main module and the barrier to have the feature is to just install the extra module. But one thing: you don't need any migration script. You just need to install the extra module. The only sequence that will lead to data loss will be to update the main module without installing the extra one.

nimarosa:
> I still don't understand why you think option 2 (move the functionality to another module and use a migration script to migrate data won't work. This is exactly the sort of thing it's supposed to do: copy the data to temporary tables, upgrade the modules, move the data from the temporary tables to the correct place.

As i understand, this is correct if you are doing it in the same module or between modules that are dependent of each other because, suppose you write the migration script in payroll module, when you update the module, the script will drop the tables, if we are going to migrate the tables to the other module, we need the other module installed first, to do that, it needs to be dependency (or at least be installed) to the new module tables exists in the first place, and then the migration script  might migrate them. But suppose the user don't manually install the module, how are you going to do to migrate this data you put in temporary tables if no module installed? 

The point is that if you today have payroll installed with parameters in the db. And you update payroll, which type of script you will need to make this work? Because the other module won't be installed yet so you won't have any table to put the data in. How the script decides if the user will install the other module later or not? maybe i'm not clear. 

Maybe i'm not getting it. I've never writed a migration script that involves two modules with no dependency on each other. If you say it's possible, great, more points to that option.

nimarosa:
> I'm also with option 2, as we don't "bloat" the main module and the barrier to have the feature is to just install the extra module. But one thing: you don't need any migration script. You just need to install the extra module. The only sequence that will lead to data loss will be to update the main module without installing the extra one.

@pedrobaeza @mtelahun @norlinhenrik

@pedrobaeza If you have a new module, and you install it, this new module will depend on payroll, so this install will trigger payroll update. The question is, what happens first? The install of the module, or the deletion of data tables from the db since payroll module will no longer have the model? That is what i have doubts about this. Also, what i think it will happen, people will update payroll and loss the data without knowing they have to install the other module first or the migration to happen. 

But if you guys think that can be solved with a migration script, let's try it:

So we are 2 vs 2 in the votes... 

In pursuit of finally getting an arrangement on this, i can change my vote to the "new module" option. We can move with that approach, as i said, i have no problem with any option only having to adapt my clients/modules but no problem. I just hope that is the correct decision for other users data integrity. 

@norlinhenrik since you made this PR, do you want to move all this implementation to a new module? It should not be difficult since you already have all new files in this PR. Then we can test and work in the migration script i can help you in what you need, but i'm not sure how to do it in the right way.

pedrobaeza:
> If you have a new module, and you install it, this new module will depend on payroll, so this install will trigger payroll update. The question is, what happens first?

This is not true. Installing a module doesn't update the dependent one if that one is already installed.

nimarosa:
> This is not true. Installing a module doesn't update the dependent one if that one is already installed.

Great to know. I was sure it was the other way. Maybe i was confusing with the inverse process when you update a module that is a dependency of another and that triggers the update of the other. 

But if it work like you said, then we can use the current migration script and adapt it to the new module so the only thing that should happen is to install the new one before updating payroll.

nimarosa (changes requested):
As discussed, we should move this functionality to a new module here in payroll repository. 
We should move new files and tests to new module, and work in the migration script to prevent data loss on current salary_rule_parameter implementation. 

Since we are moving this to a new module, adding the method in payslip object will be no longer possible. So i suggest: 

1. we use the "payroll" (inheritable object) object to store all parameters values so they become available in salary rules. 
2. We add a new object to the dict that allows us to access this parameters from the rules. 

@norlinhenrik we can work in this PR together if you need some help, i can make PR to your branch.

norlinhenrik:
I agree that `payroll` should be a minimalistic base module, and I will make a new `payroll_time_parameter` module.

pre-migration.py will execute BEFORE hr.rule.parameter(.value) will be deleted. I will [rename_table](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L730) to avoid data loss, even if `base_time_parameter` is not installed before updating `payroll`. (But there would be some manual work to move the data to base_time_parameter(_version) tables.)

pedrobaeza:
If you do that, then you can put a pre_install hook on the new module `payroll_time_parameter` that checks if the table with the renamed name is there, and if so, rename back to the proper name.

nimarosa:
@norlinhenrik great approach. Let me know if you need help. 

Can I suggest the module name to be payroll_salary_rule_time_parameter or to make it more short payroll_rule_time_parameter. 

Thanks

norlinhenrik:
@nimarosa Would you like to fork [norlinhenrik/payroll](https://github.com/norlinhenrik/payroll/tree/14.0-payroll-depends-on-base_time_parameter) and make a PR with these changes to my branch? Then when I approve it will be added to this PR.

nimarosa:
@norlinhenrik let's do this, you can start moving everything to a new module in your branch and work in the migration script (because I think you will have more experience with that type of scripts) and then I can fork and do the part of integrating the parameters to be accessible in salary rules, I have an idea to do it. 

So you just create the new module and move all the integration and you've done, adapt scripts and then I make you a PR with the salary rules part. 

What do you think?

norlinhenrik:
All new changes are in a new commit.

I have tested updating `payroll` then installing `payroll_rule_time_parameter`. The parameters are successfully migrated.

nimarosa:
@norlinhenrik great Henrik, I will to try to make the PR to your branch with the changes today, or tomorrow. Thanks.

nimarosa (approved):

nimarosa:
@norlinhenrik PR is ready for review and testing. 

Now time parameters can be accessed from salary rules doing: 

`result = payslip.get_time_parameter("CODE") # Optionally we can pass date field` 

Maybe then i can make some more changes to auto-pass date, for now, it can be accessed using the main implementation of base_time_parameter module. 

To do this, i had to change what i think was a bug. "payslip" localdict object, was referring to Payslips browsableobject object, not to the current payslip object. Now, we have a localdict "payslips" object (referring to Payslips browsableobject) and a localdict "payslip" object (referring to current payslip object). This way, we can access all payslip object functionalities from salary rules, like this time_parameter one. 

Will change the help text and documentation in next commit to reflect that change.

payroll/models/hr_payslip.py
```diff
@@ -511,7 +511,7 @@ def _get_baselocaldict(self, contracts):
             line.code: line for line in self.input_line_ids if line.code
         }
         localdict = {
-            "payslip": Payslips(self.employee_id.id, self, self.env),
+            "payslips": Payslips(self.employee_id.id, self, self.env),
```
mtelahun, line 514: Is this by mistake? Because if it isn't it will break people's rules who might be using it.

payroll/models/hr_payslip.py
```diff
@@ -618,7 +618,10 @@ def get_lines_dict(self):
                 )
                 # set up localdict with current contract and employee values
                 localdict = dict(
-                    baselocaldict, employee=contract.employee_id, contract=contract
+                    baselocaldict,
+                    employee=contract.employee_id,
+                    contract=contract,
+                    payslip=payslip,
```
mtelahun, line 624: Why are you adding payslip=payslip here? It's already provided by baselocaldict. See comment above.

payroll_rule_time_parameter/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "Payroll Rule Time Parameter",
+    "summary": "",
+    "author": "nimarosa, appstogrow, Odoo Community Association (OCA)",
+    "category": "Human Resources",
```
mtelahun: This should probably be "Payroll"

mtelahun (changes requested):
There are a couple of changes in the payroll module that don't seem to be directly related to the new module. Please have a look at them.
Thanks.

payroll/models/hr_payslip.py
nimarosa: Actually i discovered this when doing this PR and i think is a bug. Since the "payslip" object, should refer to payslip object, not to the browsable version of it (see Payslips in base_browsable.py). 

So it was never the payslip object, it was the Payslips object with "payslip" name.

nimarosa: See below answers to understand more.

mtelahun:
On 08/11/2022 08:04, Nicolas Rodriguez Sande wrote:
> To do this, i had to change what i think was a bug. "payslip" 
> localdict object, was referring to Payslips browsableobject object, 
> not to the current payslip object. Now, we have a localdict "payslips" 
> object (referring to Payslips browsableobject) and a localdict 
> "payslip" object (referring to current payslip object). This way, we 
> can access all payslip object functionalities from salary rules, like 
> this time_parameter one.


I saw this in the PR and thought maybe unrelated changes crept into the 
diff. The "payslip" object (now "payslips") should contain the current 
payslip object, if I'm not mistaken. Maybe it was inadvertently changed 
recently.

payroll/models/hr_payslip.py
nimarosa: That's the whole reason, "payslip" object was never in the localdict, it was the Payslips (browsable class) object that was in it. 

In the practice, you can access the same fields, with the difference that it's not the real payslip object so if you want to call any not stored field from the previous "payslip" object, you can't. Now with this change, you can, and also continue having the same fields so we are no breaking any rules. 

Now we have a new object called "payslips" that refer to the browsable class. But "payslip" object was never the real object. I realised now, and i think it was a bug that has passed uncaught because the browsable class also retrieves the stored values so, in the most of the cases, you didn't realize that you are no using the real payslip object. 

I tested and it has no effect in rules using "payslip".

payroll_rule_time_parameter/__manifest__.py
nimarosa: Yes you are right. Changing it in next commit.

nimarosa:
> On 08/11/2022 08:04, Nicolas Rodriguez Sande wrote: To do this, i had to change what i think was a bug. "payslip" localdict object, was referring to Payslips browsableobject object, not to the current payslip object. Now, we have a localdict "payslips" object (referring to Payslips browsableobject) and a localdict "payslip" object (referring to current payslip object). This way, we can access all payslip object functionalities from salary rules, like this time_parameter one.
> I saw this in the PR and thought maybe unrelated changes crept into the diff. The "payslip" object (now "payslips") should contain the current payslip object, if I'm not mistaken. Maybe it was inadvertently changed recently.

@mtelahun 

I checked and it's a thing that comes from the Odoo S.A original module. So it was a bug that passed uncaught because "payslip" object (the browsable one) also return all payslip stored values, so it's practically the same object with the difference that it's not the real one so i realized it was wrong when working on this PR and tried to call a method from the object. 

That's why i changed the old "payslip" object to "payslips" (like it must be from the start) and created a new object "payslip" that now contains the real payslip object, not the browsable one. 

I explained it a bit in the comments hope i made it clear. But i agree with you, "payslip" object should always have returned real payslip object, not the browsable one. The browsable one belongs to payslips, so we can use the methods (like sum). 

**To make it clear, it should not affect any rule using it because the new payslip object contains all values that the old one had, plus, it's the real payslip object and can be used now like we use contract and employee objects. Prior to that, it was a browsable class so it only returned stored values (like dates, etc).**

Edit: If you check the 13.0 branch of the module, you can see that the bug comes from long ago. We didn't notice because it's almost the same values returned.

mtelahun:
@nimarosa Ok, I understand.

We should find some other way of exposing internal objects. I don't like that employee, contract, and now payslip objects can be accessed (and potentially modified) from salary rules. But that's an idea for another PR and doesn't belong here.

norlinhenrik:
> I don't like that employee, contract, and now payslip objects can be accessed (and potentially modified) from salary rules.

Then I think we should not add the payslip object to localdict. In the browsable class can add "get_" to time_parameter and use `result = payslip.get_time_parameter("CODE")`. Currently, the date is take from `date_to`. Do we sometimes need `date_from`? If so, then instead of get_time_parameter we could have two methods: `start_date_parameter` and `end_date_parameter`.

Is there a way to inherit payroll browsable class in `payroll_rule_time_parameter`? Otherwise these methods must be inside `payroll`. That's the only downside I can see.

nimarosa:
> Then I think we should not add the payslip object to localdict. In the browsable class can add "get_" to time_parameter and use `result = payslip.get_time_parameter("CODE")`. Currently, the date is take from `date_to`. Do we sometimes need `date_from`? If so, then instead of get_time_parameter we could have two methods: `start_date_parameter` and `end_date_parameter`.

@norlinhenrik 

Despite of this module, I think that the main idea of the payslip object was being in the localdict. It was a bug that it was not. The browsable class work, but it didn't give you access to all model field and methods like contract and employee do. 
So the change it's more about fixing it and not only for that module. 

Also this not only allows to use this method, it also solves other problems like allowing accessing computed fields and methods inside payslip class, which was not possible with previous implementation. 

@mtelahun like you said, I didn't like either that the objects are modificable but I think adding payslip it's the same because we already have contract, employee and also we have the env object: even if we don't have these objects there, you can do something like:

contract.env["hr.payslip"] and there you have the payslip object again. So, like you said, I think that for now the right path is adding it because it was intended to be there from the beginning. Payslips browsable object was intended to be used to interact with other payslips not to be used as main payslip object. 

We should work in the future (maybe after migrations) in the idea of adding more standard functionalities for payslip calculation so we can make Python code an optional module, work more closely in don't allowing write operations in the code or write our own scripting language for rules. 

> Is there a way to inherit payroll browsable class in `payroll_rule_time_parameter`? Otherwise these methods must be inside `payroll`. That's the only downside I can see.

This is the whole problem. It's not possible to inherit, at least I don't see how, the browsable classes, and I'm not a fan of put code in payroll that will only be used if other module is installed. I think that's worse solution that having a dependency.

norlinhenrik:
@nimarosa has improved **help texts** in views. If there will be no new payslip object, this info should change.
The **manifest** has category Payroll instead of Human Resources. I think all modules should change this at the same time.
It is good to remove the deprecated **rule_parameter** from the browsable payslip and tests.
The other changes I suggest not to keep.

norlinhenrik:
@mtelahun Can we inherit the payroll browsable class in `payroll_rule_time_parameter`? If not, do you see that as a problem?

@nimarosa localdict without records is much safer. Then `contract.env["hr.payslip"]` will not be possible.

nimarosa:
> @nimarosa localdict without records is much safer. Then `contract.env["hr.payslip"]` will not be possible.

The problem is that we already have a bunch of records. You can edit contract and employee if you want and if you want to access payslip you can with env: 

I don't test the contract.env. The one I tested and know it can be done because I used it with prior impmemntation is:

payslip.env["hr.payslip"] 

Test it, try to call a function or field from that object and it will work. 

Edit: actually it will not work from contract because it's an object but it will work from any browsable class object so you can also do worked_days.env["model"] if you want. 

So I think the security issue it's not in adding payslip as an object because it can be accessed anyways, we have to work in safe_eval security or in writing more standard methods to move python to another module like you suggested, or even write own rule language. But this is out of scope of this PR, and adding payslip is actually the same in security concerns, it does not add security but didn't increase current risk, we already have objects and environment available in rules. If someone want to maliciously do something with a rule, they will can. That's why we added security groups and probably I will make more PRs trying to improve security. But not adding payslip literally don't change nothing to current security issues. I think @mtelahun will agree on this since it's a known problem and fixing this object (like it was intended to be) will not change it.

norlinhenrik:
> I don't like that employee, contract, and now payslip objects can be accessed (and potentially modified) from salary rules.

@mtelahun 
**If we can call** `browsableobject.env["model"]`, **then what is the security benefit of not passing records directly?**

@nimarosa It is easy to override the localdict. In `payroll_payslip_line_manually` I add the rule record:

```
    def _satisfy_condition_python(self, localdict):
        localdict["rule"] = self
        return super(HrSalaryRule, self)._satisfy_condition_python(localdict)

    def _compute_rule_code(self, localdict):
        localdict["rule"] = self
        return super(HrSalaryRule, self)._compute_rule_code(localdict)
```

nimarosa:
> > I don't like that employee, contract, and now payslip objects can be accessed (and potentially modified) from salary rules.
> 
> 
> 
> @mtelahun If we can call `browsableobject.env["model"]`, then what is the security benefit of not passing records directly?

That's the point. The security issues come from safe_eval itself not from adding a record directly. We do it with contract and employee and also have access to all env objects so, why not to add payslip object since we always thought it was there. I didn't realize that the payslip object was a browsable object until I had to call a method. The other thing is they Payslips browsable class object should not be used as payslip object, instead it's there to interact with other payslips (like the sum method does). 

> @nimarosa It is easy to override the localdict. In `payroll_payslip_line_manually` I add the rule record:
> 
> 
> 
> ```
> 
>     def _satisfy_condition_python(self, localdict):
> 
>         localdict["rule"] = self
> 
>         return super(HrSalaryRule, self)._satisfy_condition_python(localdict)
> 
> 
> 
>     def _compute_rule_code(self, localdict):
> 
>         localdict["rule"] = self
> 
>         return super(HrSalaryRule, self)._compute_rule_code(localdict)
> 
> ```

Yes I know. It was my first approach until i realized that with base_time_parameter I can call the function directly from any model. Then i realized that the payslip object was not really the real one. 

But this changes are not only motivated by this module (it help a lot because we didn't have to add any code for base_time_parameter to work) but also because accessing the real payslip object allow us to call methods inside it, that return information, like we can do with contract and employee. If we use the browsable version of payslip, we can only access stored fields, not even computed ones. So that's a limitation that is not justified by security issues because the current python implementation has already security issues because of python safe_eval, so I continue thinking that **for now** the best is to add the correct object like I'm doing. 

If then we want to work in security we have to do one of the things I listed earlier, since not adding an object now will not make the system more secure.

nimarosa:
@norlinhenrik @mtelahun i created a new issue to start working in a new language or framework to write rules that allow us to get rid of safe_eval and security concerns. 

In the meantime, like I said, I think that we should finish this PR and let payslip object exists since we have no other option for now and the system is already insecure since it has env access. Adding or removing objects will not change that. We should work on safe eval which is really the problem. 

Regards.

mtelahun:
Yes, it's a difficult problem. That's one of the reasons why I said it 
should be a problem for another PR. That's also why I don't really 
object to payslip also being directly available now. It's a hard problem 
that needs to be solved, but it's going to take time.

norlinhenrik:
Good, then `localdict["payslip"]` will be the record instead of the browsable object. Is there any behavior of the browsable object that the record does not have? If not, then why add the browsable object as `localdict["payslips"]`?

@nimarosa In the commit **[14.0] [IMP] payroll: change manifest category** could you please update the category in ALL modules in the repo?

I see that you could update my branch directly, without any PR. How did you do it?

nimarosa:
@norlinhenrik 

> Good, then `localdict["payslip"]` will be the record instead of the browsable object. Is there any behavior of the browsable object that the record does not have? If not, then why add the browsable object as `localdict["payslips"]`?

"payslips" browsable object, contains a "sum" method, that allow to sum payslip lines totals from other confirmed payslips between a date range. You should pass the line code and at least a from_date, and optionally you can pass to_date. This is useful when you need to operate with other payslips data. It use this method when i need to sum yearly gross for example, or make some average salary calculations from some rule. More useful methods could be added to this browsable object, i have plan to add an average one, and a yearly gross one. So i think it's useful to keep it. 

So now we will have two objects: 
- payslip: real original payslip object
- payslips: the browsable one, in which we can plug in methods to work with all payslips data, not only the current one. For now, we have the sum method. 

> @nimarosa In the commit **[14.0] [IMP] payroll: change manifest category** could you please update the category in ALL modules in the repo?

Yes will do in next commit. 

> I see that you could update my branch directly, without any PR. How did you do it?

When you create the PR, by default you have an option that says "Allow edits from maintainers". When this setting its activated, it allow more than you to collaborate in your own PR. This is very useful for collaboration in large PRs or features. 

But i don't know if anyone can do it. I think i can because i have write access to the repository because the PSC role i have assigned, didn't tried it in other repos (will do and let you know). 

I also do it with vscode github extension that let me select the pull request and change to the "pull request branch" and then i make the changes and push. It looks like this: 
Screen Shot 2022-11-08 at 17 34 21

So i think it's like a virtual branch that github creates and the replicates in yours. I don't know if requires some specific commands because i always do it with vscode.

nimarosa (approved):

norlinhenrik:
Thanks for the explanations.
I suggest you add this to the payroll migration script (not the best scripting but I think it works):

```
def localdict_convert_payslip_to_payslips(env):
    for rule in env["hr.salary.rule"].sudo().search([]):
        if 'localdict["payslip"]' in rule.condition_python:
            rule.condition_python = rule.condition_python.replace(
                'localdict["payslip"]', 'localdict["payslips"]'
            )
        if "localdict['payslip']" in rule.condition_python:
            rule.condition_python = rule.condition_python.replace(
                "localdict['payslip']", "localdict['payslips']"
            )
        if 'localdict["payslip"]' in rule.amount_python_compute:
            rule.amount_python_compute = rule.amount_python_compute.replace(
                'localdict["payslip"]', 'localdict["payslips"]'
            )
        if "localdict['payslip']" in rule.amount_python_compute:
            rule.amount_python_compute = rule.amount_python_compute.replace(
                "localdict['payslip']", "localdict['payslips']"
            )
```

nimarosa:
> def localdict_convert_payslip_to_payslips(env):
>     for rule in env["hr.salary.rule"].sudo().search([]):
>         if 'localdict["payslip"]' in rule.condition_python:
>             rule.condition_python = rule.condition_python.replace(
>                 'localdict["payslip"]', 'localdict["payslips"]'
>             )
>         if "localdict['payslip']" in rule.condition_python:
>             rule.condition_python = rule.condition_python.replace(
>                 "localdict['payslip']", "localdict['payslips']"
>             )
>         if 'localdict["payslip"]' in rule.amount_python_compute:
>             rule.amount_python_compute = rule.amount_python_compute.replace(
>                 'localdict["payslip"]', 'localdict["payslips"]'
>             )
>         if "localdict['payslip']" in rule.amount_python_compute:
>             rule.amount_python_compute = rule.amount_python_compute.replace(
>                 "localdict['payslip']", "localdict['payslips']"
>             )

Actually we should replace payslip with payslips. It's not necessary to call localdict inside rule, you call directly the object. I will adapt the code and add it to the migration in case this change affects anyone. 

Edit: @norlinhenrik done. Please check last commit, i'm not so experimented with migration.

norlinhenrik:
Usually all methods are inside the migration script. I put move_records outside of (14.0.6.0.0) so I could import it also in a hook.

In pre-migration, the models are not loaded yet, so we cannot call `env["hr.salary.rule"]` in pre-migration. So create a new file post-migration.py:

```
from openupgradelib import openupgrade


def localdict_convert_payslip_to_payslips(env):
    for rule in env["hr.salary.rule"].sudo().search([]):
        if "payslip" in rule.condition_python:
            rule.condition_python = rule.condition_python.replace("payslip", "payslips")
        if "payslip" in rule.amount_python_compute:
            rule.amount_python_compute = rule.amount_python_compute.replace(
                "payslip", "payslips"
            )


@openupgrade.migrate()
def migrate(env, version):
    localdict_convert_payslip_to_payslips(env)

```
I have tested that this code works.

nimarosa:
@norlinhenrik Done now. I think... please check it and thanks for the input.

norlinhenrik:
post-migration works.
But `employee.payslip_count` should not become `employee.payslips_count`.
So please replace "payslip." to "payslips."
This would still change `env["hr.payslip.line"]` to `env["hr.payslips.line"]`
So maybe it is better to drop this migration and manually change payslip to payslips when needed?

nimarosa:
> post-migration works.
> 
> But `employee.payslip_count` should not become `employee.payslips_count`.
> 
> So please replace "payslip." to "payslips."
> 
> This would still change `env["hr.payslip.line"]` to `env["hr.payslips.line"]`
> 
> So maybe it is better to drop this migration and manually change payslip to payslips when needed?

Yea I agree. Anyways if the user is using payslip (the old one) to access payslip fields it will continue working after the change. The only case the user must change it manually is if they are using the sum method of the browsable, so I think it's no a problem to leave that migration out. 

Will make the changes and finally merge so we move forward.

nimarosa (approved):

nimarosa:
Okay so migration is removed, everything else is working. I think we can merge this one. 

/ocabot merge major

PR #88, opened 2022-10-17
[14.0][ADD] payroll_salary_rule_python: removing from "payroll"

norlinhenrik:
This PR is moving user-defined python code from `payroll` to `hr_salary_rule_python`.

The reason is that a payroll user who may edit python code may edit ANYTHING by using the database cursor. It requires the same level of trust as a database manager. So I would like to encourage implementations that are not using user-defined python code.

In comparison, user-defined python code for taxes is not in `account`, but in `account_tax_python`.

I have not updated the tests yet.

@mtelahun and @nimarosa, what do you think?


mtelahun:
I like the idea. As I've said multiple times I am not in favor of users 
running full-blown python code in salary rules. I haven't looked at the 
code yet, but I have one suggestion. I think 
``payroll_salary_rule_python`` would be a better name for the module.

norlinhenrik:
This PR is rebased and ready.

**NB!** Custom modules depending on `payroll` should maybe instead depend on `payroll_salary_rule_python`.

nimarosa:
Hello, i had a very busy week but will try to catch up with the new PRs this following days.

As i understand, you are removing all python code to a new module. But what this solves? If the user installs the module i will have to install this new one usually in any use case. I don't quite see the benefit of it. I understand the problem of running python code introduced by the user, but i think we can find another solution because as payroll module is today, running python code will be needed almost always because the only other options are Fixed Amount, or Percentage and Range in conditions. This two cases don't cover a lot of common cases that we should need in making salary rules. 

I think that if we want to remove python code from rules to a new module, first we need to work in introducing more amount_select and condition_select options, so then we can encourage the user to use standard calculation options. 

In resume, it seems like a good idea but i think first we need to create more standard options, because the module will be affected if we remove this, because new users of the module might don't know about this separate module and judge the module by the standard options listed. 
Also i want to note that in all modern payroll systems it's allowed to write rules, that's because payroll is very different in every use case, country or law, so we should be able to write code. 

I propose the following solutions, at least until the time we have more standard options that allows the user to use payroll without writing python salary rules:

- We create a setting that locks/unlocks the python code form, this setting will be only possible to be changed by an user with administrator permissions (we can limit this option to developer mode or something like this). So when the setting is disabled, all forms will show the code, but will not allow editing it or changing anything of it. That way, we gain more security without having to remove the functionality yet. 
- We can work in a serie of PRs regarding adding more "standard" options to condition_select and amount_select, so the user can use payroll without having to write code. 
- I will research a way of making safe_eval more secure, maybe there is a way of limiting database cursor or something like that.

ghost:
> We create a setting that locks/unlocks the python code form

Good idea.

> adding more "standard" options to condition_select and amount_select

Also good idea.

> making safe_eval more secure

I have modified safe_eval() [here](https://github.com/appstogrow/OCB/blob/14.0-mbase/odoo/tools/safe_eval.py#L329).

(A) Should we add `payroll_salary_rule_python` now? (Maybe skip 14.0, but add it in 15.0 and 16.0?)

(B) Should we first develop more "standard" options, and add `payroll_salary_rule_python` next year?

(C) Should we never never add `payroll_salary_rule_python`, just create a setting that locks/unlocks the python code form?

I think (A) we can implement `payroll_salary_rule_python` soon and add "standard" options later, and uninstall `payroll_salary_rule_python` when it is not needed anymore. But it is not urgent for me, and I value your opinions.

nimarosa:
@appstogrow Hello Henrik, in my opinion I think that we should add the setting in this PR, and then in the future start developing more standard functions. When this is done, I think we could add this separate module and see how the people adopts the change. 

If we have a lot of standard functions I think It's a positive change but without them, the installation of this module will be mandatory almost always so for know I think we should not remove it from payroll base module. 

We can use this PR to work in this new setting.- blocking the form will prevent the most users of editing anything in the rules, so, then we can also add some checks for when the user saves the form, meaning that if user changed anything and the setting is turned off, it will reject saving. I think this is a good solution for most cases. 

--- 

About safe_eval I think the problem is that if we change "exec" param, we will not be able to do calculations or conditions.

Edit: also a option is to block the whole form using this setting, something like an error when you try to edit that says "salary rules edition can only be enabled by administrator. Contact your administrator." Or something similar. This way we introduce the concept to the user that rules Can and should only be edited on crucial moments and that has to be allowed by some admin. Also this prevents not only the python code part, it also prevents the user to alter salary rules to change payslip computations (for example for his own salary).

nimarosa:
@norlinhenrik Hello Henrik, do you think we could finish this PR and the other one of yours this week? That way we merge and we can ask @mtelahun to integrate this changes in the migrations so we merge migrations too.

ghost:
This PR is about splitting python into a separate module.
@nimarosa Is it so that you would not like to have this in version 14.0?
@mtelahun What do you think about adding this in version 16.0?

A setting that locks/unlocks the python code should be made in a new PR. I will not prioritize that this week.

nimarosa:
@appstogrow Hello Henrik. Like we discussed in previous comments I thought that we both agree about delaying this PR until we have more standard functions to add to condition_select and amount_select because if we remove python without having more standard calculation options this module will almost always need to be installed, so if we don't have that the purpose of this PR will not be useful since in almost every possible use case the user will have to install it anyways. 

That's why according to your questions in last comment (the one with your other username) I think until we have more standard options available we can make the setting-based salary rule form block to archive the security that salary rules needs (rule edition should be only enabled by some user with high permissions) without separating the python calculation for now. 

In the future versions or PRs if we add finally add new standard calculation options we can remove the setting and separate python to a new module like this PR does. 

But I think this should only be done if we extend standard available options so the user can make complex rules without python. When that happens, i like and make sense to move python outside module. But separating python to another module now will have no benefit because the user would have to install the new module almost always making it almost a dependency of payroll. 

---

So, in resume, I think we should do this: 

1. Create the form block setting now in 14.0 to archive the desired security improvements and to prevent user to edit rules without permission and exploit the system. 
2. In future versions or in 14.0 we can work in developing standard computation options to cover the most salary rule use cases that we can. 
3. When point (2) is considered mature enough we can continue this PR and separate python code computation method from "payroll" to another module, incentivizing the users to prefer the use of standard computation methods and leaving python in separate module only for cases that really need to write python. 

So I don't know if this will happen in version 16.0 or in 14.0 I think it depends on when we can start working in the new computation options. But what can be done and also merged with the migrations is the setting that will allow us to have more security in salary rules without too much effort. Then in the future we can start working in the other changes and finally merge this PR in the main branch of the time we finish with everything. 

So I suggest making this a WIP draft PR until we finish working on prior steps/PRs

nimarosa:
About the priority, I have no rush for it so it's okay for me. Also not rush with migrations but will be good merging them maybe by the end of next month, so we don't delay migration too much.

PR #89, opened 2022-10-24
[14.0] [ADD] hr_payroll_period

nimarosa:
As discussed in oca/hr repository https://github.com/OCA/hr/pull/1124. We are moving hr_period module to payroll, since it's more related to payroll. 

I will open the PR, but want to make some modifications to the module, so expect new commits here.


nimarosa:
Merging this since it's the same module that was in oca/hr so we can migrate it to other versions like planned. 

Module don't affect current ones so I think it's safe to just merge it. So we start finishing current open PRs. 

/ocabot merge nobump

etobella:
@nimarosa You never maintained the commit history, it is right? This will be the new module name? Right now, OCA/hr is failing due to some incompatibilities between hr_period and payroll changes applied. We have 2 options:

Fix them and have two modules with the same logic or make the one from hr a dummy module that install the one from payroll. In any case, I think we need to restore original commit history here.

PR #90, opened 2022-10-25
[14.0][IMP] payroll: Only payroll manager can edit salary rules

norlinhenrik:
See discussion in #88 
What do you think about this solution?

### DIFF
--- a/payroll/security/ir.model.access.csv
+++ b/payroll/security/ir.model.access.csv
@@ -9,7 +9,8 @@ access_hr_payslip_input_user,hr.payslip.input.user,model_hr_payslip_input,payrol
 access_hr_payslip_worked_days_officer,hr.payslip.worked_days.officer,model_hr_payslip_worked_days,payroll.group_payroll_user,1,1,1,1
 access_hr_payslip_run,hr.payslip.run,model_hr_payslip_run,payroll.group_payroll_manager,1,1,1,1
 access_hr_rule_input_officer,hr.rule.input.office,model_hr_rule_input,payroll.group_payroll_user,1,1,1,1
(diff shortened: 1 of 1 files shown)


nimarosa:
Looks good solution for me. Maybe I would suggest to call the new group "Payroll Superadmin" or "Payroll Admin" since this group will have all permissions too and can be used to limit and improve the other roles in the future. 

But looks good to me. 👍

nimarosa:
@norlinhenrik Also check pre commit failing .

mtelahun:
I don't understand what the point of this new group is. Having an Officer and a Manager makes sense. You can divide responsibilities for running payroll and configuring payroll. A new group that has a Manager's permissions and has only the additional permission of writing python code just doesn't make any sense to me. It's simpler and easier just to keep it a responsibility of the payroll manager.

payroll/models/hr_payslip.py
```diff
@@ -623,9 +623,13 @@ def get_lines_dict(self):
                 for rule in payslip._get_salary_rules():
                     localdict = rule._reset_localdict_values(localdict)
                     # check if the rule can be applied
-                    if rule._satisfy_condition(localdict) and rule.id not in blacklist:
+                    rule_sudo = rule.sudo()
```
mtelahun: Why is this running with admin privileges? If anything, doesn't this make it less secure? Before it was running as Payroll Manager now it's running as Administrator.

nimarosa: I didn't saw this change. Why is this sudo necessary now?

norlinhenrik:
Ok I made it very simple now.

nimarosa (approved):

PR #91, opened 2022-10-27
[14.0] [IMP] payroll: Add employee in rule exceptions

nimarosa:
This PR adds employee name to the exception message in salary rule computation. 

This is useful mainly for the cases when you are creating a payslip batch (payslip run) and you get an error, but given that you are making payslips for more than one employee, you can't know which employee is giving the error. 

This improvement add employee name to the error so you could easily debug a payslip batch.

### DIFF
--- a/payroll/models/hr_salary_rule.py
+++ b/payroll/models/hr_salary_rule.py
@@ -208,8 +208,8 @@ class HrSalaryRule(models.Model):
             }
         except Exception:
             raise UserError(
-                _("Wrong quantity defined for salary rule %s (%s).")
-                % (self.name, self.code)
+                _("Wrong quantity defined for salary rule %s (%s) for employee %s.")
+                % (self.name, self.code, localdict["employee"].name)
             )
 
     def _compute_rule_percentage(self, localdict):
@@ -224,9 +224,9 @@ class HrSalaryRule(models.Model):
             raise UserError(
                 _(
                     "Wrong percentage base or quantity defined for salary "
-                    "rule %s (%s)."
+                    "rule %s (%s) for employee %s."
                 )
-                % (self.name, self.code)
+                % (self.name, self.code, localdict["employee"].name)
             )
 
     def _compute_rule_code(self, localdict):
@@ -237,13 +237,13 @@ class HrSalaryRule(models.Model):
             raise UserError(
                 _(
                     """
-Wrong python code defined for salary rule %s (%s).
+Wrong python code defined for salary rule %s (%s) for employee %s.
 Here is the error received:
 
 %s
 """
                 )
-                % (self.name, self.code, repr(ex))
+                % (self.name, self.code, localdict["employee"].name, repr(ex))
             )
 
     def _get_rule_dict(self, localdict):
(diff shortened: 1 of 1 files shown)


nimarosa:
@mtelahun @appstogrow Are you guys okay merging this? Please if you can review it. This weekend i would dedicate time to review all open PRs so we can merge them fast and continue with migrations. 

We can start with this one that is a simple improvement to the errors.

norlinhenrik (approved):

mtelahun:
Rather than employee name, it should be payslip name. What happens when multiple employees share the same name?

nimarosa:
> Rather than employee name, it should be payslip name. What happens when multiple employees share the same name?

@mtelahun Payslip name contains also the name of the employee and the date (if you are using default name which it's not always the case). So if we use payslip name I think it's the same because if two employees share the names payslip name will be the same too. 

Anyways it's quite a coincidence if two employees share their names, and even if they do, the reason for this change is to identify which payslip is giving you the error on a payslip run. So if you have more than one employee with that name, you can check all of them without hassle. 

The main problem before is that when you do payslip runs and receive a rule error you only see rule name and code, and the whole batch is rollbacked so you can't know which employee payslip is giving you the error. Printing employee name makes it a lot more easy even if two employees share their name. So I think it's no a problem, and employees sharing name and surname can be at least 2 o 3 cases which are also easy to check.

mtelahun:
On 02/11/2022 15:05, Nicolas Rodriguez Sande wrote:
>
>     Rather than employee name, it should be payslip name. What happens
>     when multiple employees share the same name?
>
> @mtelahun <https://github.com/mtelahun> Payslip name contains also the 
> name of the employee and the date (if you are using default name which 
> it's not always the case). So if we use payslip name I think it's the 
> same because if two employees share the names payslip name will be the 
> same too.
>
> Message ID: ***@***.***>

Sorry, I meant payslip reference number. It also occurs to me that 
having employee names hanging around in log files is not good security 
practice. You are potentially leaking sensitive information when you 
don't have to.

nimarosa:
@mtelahun yes I also wanted to use slip reference number. The problem is that this number is generated when the payslip is successfully computed. So if you get a rule error, you don't have reference number generated. 

Also, when you create and compute a payslip run, if one of the payslip involved fails, all previous payslip are rollbacked so they no longer exists. 

So the only variable that I thought we could use to identify which employee has the problem is the name. We can also use contract name but it also by default contains the name of the employee so it's the same of using employee name. 

If you can think in another way of identifying the employee without showing the name in the error please suggest it and we can work it out. But I think we have no choice here since the only way to identify the employee when you get an error is by the name.

mtelahun:
I see indeed. I'm still not happy about leaking potentially sensitive 
information in logs, but perhaps we have no other option here. I'll let 
you know if I think of anything.

nimarosa:
@mtelahun Yes I see the point and you are right. But without it is impossible to debug big payslip run. 

I have a 2000 employee client so you can imagine that when a payslip run fails it's almost impossible to know which employee failed without this. 

Maybe we can search for a way of spoofing this info in logs and show it in the error, or maybe another alternative solution. You are always more than welcome yo propose a better solution if you find one.

Thanks for your input.

norlinhenrik:
Why not use the record id? That would securely identify the employee.

`localdict["employee"].id`

Issue #94, opened 2022-11-08
Solution for safe_eval security issues


nimarosa:
Hello, motivated by the chat we were having in #87 I open this issue to discuss possible solutions to safe_eval lack of security so we can stop exposing all python environment in a salary rule. 

The main issue here is: **safe_eval is not safe at all. It exposes real environment python objects and allows the user to run almost everything from a salary rule.**

So I come up with 3 ideas that I think if we apply at least one of them it will improve security in the module rules:

1. We should write more standard options of calculations, so we can move the python mode to another module to be used only when necessary. 

2. We modify safe_eval to not allowing write access to the ORM. Actually don't know how difficult this can be, but I think is an option. 

3. We create a new salary rule coding language. It would involve a whole module change, but for me, is the more promising one, and I have some resources you can check to get an idea: 

In python we have some libraries and derivates projects called "Python String Parsers" that in resume provide a whole package that parse strings and convert them to encapsulated python code (so it didn't expose the environment) and return the result of the operation. This packages will solve the problem since we will not write python code in the rule, we will be writing strings and functions from the parser library and the the package will take care of interpreting and processing the sting code, not exposing in any way the python environment. 

For you to read I leave you some resources about it:
- [ply](http://www.dabeaz.com/ply/): Lexx/yacc library. It could provide us with a internal language from rules. In the link you can also see projects writed with ply that could be useful for providing us functions to use in rules. 
- [pyparsing](https://github.com/pyparsing/pyparsing/): it's a big module with all sort of functions and use cases. It can also provide us a string-like programming language to use in rules.

Also I leave some interesting article about eval safety issues. That is a problem with eval and safe_eval so that's what we should attack. [Eval is really dangerous](http://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html)

--- 

So in resume I think we should take a look to all options. Also, and in a way of providing temporary security I will work in @norlinhenrik approach of locking the salary rule form from only a few cases or maybe a setting. So only trusted odoo users can edit rules. 

Also I think #87 should be merged with the new payslip object fix, because adding it don't increment current security issues of salary rules, it's just another object and since we have access to all env in rules, I think we can tolerate this change. We should attack security issues from the root, which is python safe_eval not the objects. 

[...]

PR #95, opened 2022-11-08
[14.0] [IMP] payroll: improve security

nimarosa:
Hello, this commit purpose is to improve security rules and add more restriction specially to editing python rules. 

I will be making another tweaks to it in the following days, for now the changes are: 

- Created a new group, which is a hidden one (can only be activated in the user form, while being in developer mode and with a checkbox)
- Users with this group, will be the only who can see the "python code" fields. 
- Other users, including the payroll manager, will see an error message it's shown in the following screenshot: 
Screen Shot 2022-11-08 at 18 03 45

This is to make the user understand why he can't see the python code field, and don't believe it's a bug. 
I have the intention of doing some tweaks to the other existing groups, so we can have very detailed and extensive security rules and groups, that could prevent the most any exploit of python safe_eval functionality. We will achieve that by limiting the roles, specially with all python related stuff, but not limited to. 

Keep posted for more commits.

### DIFF
--- a/payroll/security/hr_payroll_security.xml
+++ b/payroll/security/hr_payroll_security.xml
@@ -25,9 +25,7 @@
         </record>
         <!-- Allow python editing -->
         <record id="group_payroll_allow_python" model="res.groups">
-            <field
-                name="name"
-            >Payroll: Allow editing python code in salary rules</field>
+            <field name="name">Payroll: Access to python code in salary rules</field>
             <field name="category_id" ref="base.module_category_hidden" />
             <field name="implied_ids" eval="[(4, ref('payroll.group_payroll_user'))]" />
         </record>
--- a/payroll/security/hr_payroll_security.xml
+++ b/payroll/security/hr_payroll_security.xml
@@ -23,6 +23,12 @@
                 eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"
             />
         </record>
+        <!-- Allow python editing -->
+        <record id="group_payroll_allow_python" model="res.groups">
+            <field name="name">Payroll: Allow to edit python code in rules</field>
+            <field name="category_id" ref="base.module_category_hidden" />
+            <field name="implied_ids" eval="[(4, ref('payroll.group_payroll_user'))]" />
+        </record>
         <record id="base.default_user" model="res.users">
             <field name="groups_id" eval="[(4,ref('payroll.group_payroll_manager'))]" />
         </record>
--- a/payroll/security/hr_payroll_security.xml
+++ b/payroll/security/hr_payroll_security.xml
@@ -25,7 +25,9 @@
         </record>
         <!-- Allow python editing -->
         <record id="group_payroll_allow_python" model="res.groups">
-            <field name="name">Payroll: Allow to edit python code in rules</field>
+            <field
+                name="name"
+            >Payroll: Allow editing python code in salary rules</field>
             <field name="category_id" ref="base.module_category_hidden" />
             <field name="implied_ids" eval="[(4, ref('payroll.group_payroll_user'))]" />
         </record>
--- a/payroll/security/ir.model.access.csv
+++ b/payroll/security/ir.model.access.csv
@@ -9,8 +9,8 @@ access_hr_payslip_input_user,hr.payslip.input.user,model_hr_payslip_input,payrol
 access_hr_payslip_worked_days_officer,hr.payslip.worked_days.officer,model_hr_payslip_worked_days,payroll.group_payroll_user,1,1,1,1
 access_hr_payslip_run,hr.payslip.run,model_hr_payslip_run,payroll.group_payroll_manager,1,1,1,1
 access_hr_rule_input_officer,hr.rule.input.office,model_hr_rule_input,payroll.group_payroll_user,1,1,1,1
-access_hr_salary_rule_user,hr.salary.rule user,model_hr_salary_rule,payroll.group_payroll_user,1,0,0,0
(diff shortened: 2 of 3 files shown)


norlinhenrik:
Wow, I didn't see this UI tweak before.

The security setting says "Payroll: Allow **editing** python code in salary rules", but it also applies for **reading**. Either let the user read the code, or change the text, e.g. "Payroll: **Show** python code in salary rules" or "Payroll: **Access to** python code in salary rules".

I think the big pink box is too much of a warning. And "You are NOT allowed" might be discouraging to the user. I would replace all the text with a simple sentence:
"For security reasons, access to [or: editing] python code is restricted."
I don't mind including the last sentence: "If you think there is an error, please ask your system admin**is**trator." (add "please" and fix "adminsitrator").

nimarosa:
@norlinhenrik 
> Wow, I didn't see this UI tweak before.

It's very good, I use it a lot to show messages. It can also be warning and info decoration.


> The security setting says "Payroll: Allow **editing** python code in salary rules", but it also applies for **reading**. Either let the user read the code, or change the text, e.g. "Payroll: **Show** python code in salary rules" or "Payroll: **Access to** python code in salary rules".

Great, yes, I was thinking about changing the tests. So gonna accept the suggestion. 

> I think the big pink box is too much of a warning. And "You are NOT allowed" might be discouraging to the user. I would replace all the text with a simple sentence:
> 
> "For security reasons, access to [or: editing] python code is restricted."
> 
> I don't mind including the last sentence: "If you think there is an error, please ask your system admin**is**trator." (add "please" and fix "adminsitrator").

Great will make the changes. I can also make the box be yellow using the warning class. What do you think? It can be blue if I change it for info decoration.

norlinhenrik:
I think this is information, not a warning.

norlinhenrik:
@mtelahun Any opinions?

nimarosa:
> I think this is information, not a warning.

Okay so a info will be. Info box it's light blue. Will change it and post screenshot to make more easy to us to decide.

mtelahun:
Hello @nimarosa,

Thank you for starting this PR and proposing a solution. However, what 
you are suggesting is what people in the security industry call 
"security through obscurity". That means trying to make something more 
secure by hiding it. This is a practice that is not recommended and is 
discouraged as a security measure because it doesn't provided any 
meaningful security. With respect to payroll rules the security problem 
we have is not that it is easy to see or edit. The problem is that it 
allows salary rules to run python code against the odoo ORM without any 
restrictions. Simply making administrators run through hoops to 
administer their system just pisses them off and makes it harder for 
them to do their jobs without providing any meaningful obstacle to 
adversaries. So, I am not in favor of this type of security. As I 
mentioned before, short of introducing a domain specific language for 
payroll rules I don't know if there is any easy solution. This requires 
some careful thought and consideration and possibly some security people 
we can bounce our ideas off of.


Regards,

Mike.

mtelahun:
> @mtelahun Any opinions?

Hi @apps2grow ,
Please see my reply to @nimarosa. I appreciate the enthusiasm and the willingness to tackle this problem, but it will require careful consideration and the input of security people if we want to do it right.

nimarosa:
@mtelahun yes, I know this don't fix the security problem which is safe_eval like. I think the only solution is get rid of it writing a new language like I stated in issue #94. It will be great to start discussing it there so we can start working in it in a new branch dedicated to build that feature. 

About this PR, I understand your point of view but I really think limiting the access to only specific users allowed to write python is indeed good for reducing the risks. 

Of course, an experimented hacker will always find the way to hack the system, because safe_eval is not safe at all. But this change will prevent a normal user at least to edit the rules to make harmful changes to the system. I see it not as a solution but a palliative. 

For now, the only way to make safe_eval safe, is to trust user input, so I think limiting the functionality to only specific allowed users (and not the whole payroll manager group which can be composed by lot of users). With this change, you can specifically allow only trusted users to edit rules, or even you can allow it only for making changes and disallowing it after.

I know it's not really a security improvement but I think is positive to have it, even if it's requires more work from system administrator. The reality is rules don't change every day, so when all rules are created it's you will only change it if you need to adapt it to a new case o change something in it. 

Also this continue allowing payroll managers to use and edit the standard functions in rules, it only limit python fields. 

Finally, the PR is not finished but apart of hiding the field I'm planning of checking security group also in save and update methods, so if anyone manages to see the field without having the security group, it will not be allowed to save the form anyways. 

Finally, I think it's not too much hassle to turn on a checkbox in the user to allow the user to edit the rule. How many people really in a payroll or hr department edit rules on daily basis? In my experience it's only a few people in the businesses that create and code rules, so it makes sense to not allow it to all "normal" users. 

Like I said, it's not perfect and I accept more suggestions to this PR. But I think is necessary to implement something like this at least until we can replace safe_eval completely, but it's a large project and we should provide more security to payroll users.

nimarosa:
The whole point of this is not fighting an experienced hacker. An experienced hacker will always found a way to crack the system and also not only in this module. We have other models in odoo that allow to write code with safe_eval. 

The purpose of it is to reduce the amount of possible treats from normal users, or from someone in the business who know a bit of python and can exploit it. 

Like I said, it didn't add security, but I think it reduce the chances that every payroll user could exploit it. For normal users, if they can't see it or edit it, they will desist in exploiting it. So I think it's more secure that allowing all payroll managers security group to access it.

nimarosa:
Changes in last commits: 

- Accepted suggestion by @norlinhenrik. Now the "error" is a informational one, and made some tweaks to make it look good.

Now it looks like this: 

Screen Shot 2022-11-09 at 12 52 09

@norlinhenrik What do you think?

mamcode (approved):
Please make squash of commits

PR #101, opened 2022-11-16
[16.0] [MIG] payroll_hr_public_holidays: Migration to version 16.0

nimarosa:
Standard migration.


nimarosa:
**This PR has to wait the migration of oca/hr_holidays_public to be merged.**

nimarosa:
Will open new PRs with new migrations.

peterromao:
Hello nimarosa,
I am not very good with this commits and PR's but I went through the code and there doesn't seem to be any reason why this shouldn't be in the 16.0 lineup of modules.
Is only an administrative thing needed to push this?
I would help just don't know how. Most I can do is setup a local odoo install and mock up a 16.0 version of this module and come back here and say it works or not.
Thank you

PR #106, opened 2022-11-17
[14.0] [FIX] payroll_rule_time_parameters: Add rule_parameter method

nimarosa:
Here we add the rule_parameter method to the new parameters module for two reasons: 

- Mantain old rules calling rule_parameter from the code. I had problem with people getting used to the new sintax so for now the method will remain the same in payslip, and call get_time_parameter from inside.
- In salary rules we almost always need float values. I don't know if you @norlinhenrik made the value field as text for any of your requirements, so for now, we default to float only in payroll implementation. Maybe in the future we can improve this in base_time_parameter. 

Regards.

### DIFF
--- a/payroll_rule_time_parameter/__init__.py
+++ b/payroll_rule_time_parameter/__init__.py
@@ -1,1 +1,2 @@
 from .hooks import pre_init_hook
+from . import models
--- a/payroll_rule_time_parameter/models/__init__.py
+++ b/payroll_rule_time_parameter/models/__init__.py
@@ -1,0 +1,4 @@
+# Copyright (C) 2021 Nimarosa (Nicolas Rodriguez) (<nicolasrsande@gmail.com>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import hr_payslip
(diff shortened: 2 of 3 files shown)


norlinhenrik (approved):
I have not tested, but the code looks good.

So far I am not using value_text, only value_reference.

norlinhenrik:
> Maybe in the future we can improve this in base_time_parameter.

I will do it now.

nimarosa:
> > Maybe in the future we can improve this in base_time_parameter.
> 
> 
> 
> I will do it now.

@norlinhenrik okay so if you don't need text I think the field must be a float. I only store floats there but don't know if it's useful for other things to have a text field. 

Also can you explain me how you use value_reference? I tried to figure it out but how do you populate the drop-down?

norlinhenrik:
It is a bit complicated. See [here](https://github.com/appstogrow/apps/blob/14.0/account_statement_excel_import/models/base_time_parameter_version.py). Just added it to readme.

nimarosa:
> It is a bit complicated. See [here](https://github.com/appstogrow/apps/blob/14.0/account_statement_excel_import/models/base_time_parameter_version.py). Just added it to readme.

I like it, maybe i will use it for another module that might be helpful. Thanks.

Merging this since the function is required to maintain compatibility with salary rules created for the old model and the I open another one to change the float part when we merge en server-tools. 

/ocabot merge patch

PR #113, opened 2022-12-22
[16.0][FW] [14.0] [IMP] payroll: Add new "payslips" object methods

nimarosa:
Port of #109 from 14.0 to 16.0.


nimarosa:
Will FW this PR with others in the future.

PR #114, opened 2022-12-24
[IMP][14.0] hr_payroll_period: use hr_period_id.name in _compute_name()

hitrosol:
SInce hr_payroll_period has hr_period_id, therefore it's good to use hr_period_id.name as name of each payslip.

### DIFF
--- a/hr_payroll_period/models/hr_payslip.py
+++ b/hr_payroll_period/models/hr_payslip.py
@@ -79,3 +79,10 @@ class HrPayslip(models.Model):
         elif vals.get("date_to") and not vals.get("date_payment"):
             vals["date_payment"] = vals["date_to"]
         return super(HrPayslip, self).create(vals)
+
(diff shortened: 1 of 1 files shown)


nimarosa:
@hitrosol The problem with this is that if we do this, we drop support for customs modules who can be inheriting or overriding the names. Also personally I think the name is too long for a payslip if we add the period. 

But if it's a functionality you or other users want, I suggest to: 

- Make a setting in Payroll Settings to enable/disable using the period in naming. 
- Maybe we should contemplate the case when payslip name has been modified by the user or just append the period to the name without overriding the full name, something like: 

self.name = self.name + period_name 

That way we append the period instead of modifying the full name that could also be modified by another modules. 

I think the more important of the two is making a setting to let the user choose.

nimarosa:
Also I think we need to call super() in the function so we don't override others inheriting it.

PR #117, opened 2023-01-23
[14.0][IMP] payroll: salary rule code: payslip.line_sum_where()

norlinhenrik:
payslip.line_sum_where() may be used in salary rule code.
It will sum the total of the previously computed rules where the given field has the given value.
`total_seq_10 = payslip.line_sum_where("sequence", 10, rules, result_rules)`

I have also improved the salary rule help text a little bit.

### DIFF
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -804,3 +804,18 @@ class HrPayslip(models.Model):
             return line[0].total
         else:
             return 0.0
+
+    def line_sum_where(self, field_name, value, rules, result_rules):
+        """
+        The method may be used in salary rule code.
+        It will sum the total of the previously computed rules
+        where the given field has the given value.
+        E.g.: total_seq_10 = payslip.line_sum_where("sequence", 10, rules, result_rules)
+        """
+        return sum(
+            [
+                result_rules.dict[code].dict["total"]
+                for code, rule in rules.dict.items()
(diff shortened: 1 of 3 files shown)


nimarosa:
@norlinhenrik Hi Henrik, looks good but i think we should add a date_from, date_to filter to it, because this will sum all rules in history and maybe that's not the expected behaviour. 

I think in the most common use cases you will want to sum rules between two dates, or at least, beginning from a date_start param to now. Is this the case? If not, what is the use of sum all rules in history, we will end up summing years of payslip rules if we don't scope this to a date.

nimarosa:
@norlinhenrik I deleted my previous comment since i misunderstood the functionality. To be clear, this sum only the rules in the "current" payslip, not previously computed rules from another payslips. I'm right?

norlinhenrik:
Yes, it sums only computed rules for the "current" payslip.

nimarosa:
@norlinhenrik 
Okay great for me. We add it. 

If you can make some test of the new method so we maintain compatibility in future versiones will be great. 

After that, we merge.

nimarosa (approved):

norlinhenrik:
@nimarosa 
I have made a test now for payslip.line_sum_where().
I also changed so that localhost will get values by rule.id if rule.code does not exist. The tests are passing. :)

nimarosa (approved):

PR #124, opened 2023-04-19
[15.0][FIX] payroll: Change mail template expressions

victoralmau:
Change mail template expressions according to 15.0

@Tecnativa

### DIFF
--- a/payroll/__manifest__.py
+++ b/payroll/__manifest__.py
@@ -2,7 +2,7 @@
 
 {
     "name": "Payroll",
-    "version": "15.0.1.1.1",
+    "version": "15.0.1.2.0",
     "category": "Payroll",
     "website": "https://github.com/OCA/payroll",
     "sequence": 38,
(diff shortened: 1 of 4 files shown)


nimarosa:
@victoralmau the fix is okay we can merge it but please don't bump the version manually, it will be done by the bot when I merge the PR. 

Please change that and we can merge.

chienandalu (approved):
:+1:

payroll/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 
 {
     "name": "Payroll",
-    "version": "15.0.1.1.1",
+    "version": "15.0.1.2.0",
```
nimarosa, line 5: Please do not update the version. The bot will do it automatically.

nimarosa (changes requested):

payroll/__manifest__.py
pedrobaeza: The manual bump is correct, as there are migration scripts joined to that version, and launching incorrectly the bot command will make them useless. Please remove your review and merge with `nobump`.

nimarosa: Yes but I planned to use the "minor" option of the bot to bump to the correct version. 

But okay, we do it like you said.

nimarosa (approved):

PR #125, opened 2023-04-20
[16.0][FIX] payroll: Change mail template expressions

victoralmau:
FWP from 15.0: https://github.com/OCA/payroll/pull/124

Change mail template expressions according to 15.0

@Tecnativa

### DIFF
--- a/payroll/__manifest__.py
+++ b/payroll/__manifest__.py
@@ -2,7 +2,7 @@
 
 {
     "name": "Payroll",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
     "category": "Payroll",
     "website": "https://github.com/OCA/payroll",
     "sequence": 38,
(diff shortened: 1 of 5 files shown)


payroll/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 
 {
     "name": "Payroll",
-    "version": "16.0.1.0.0",
+    "version": "16.0.2.0.0",
```
pedrobaeza: I think such high jump is not needed for this fix. I know that contributing guidelines expresses that it should be major when migrations scripts are needed, but it's referring to huge database model changes, not when we need to load an XML.

pedrobaeza:
Check CI

pedrobaeza (approved):

nimarosa (approved):

PR #144, opened 2024-02-15
[16.0][ADD] payroll_fetch_prev_month

norlinhenrik:


payroll/models/res_config_settings.py
```diff
@@ -35,3 +35,9 @@ class ResConfigSettings(models.TransientModel):
         help="Require rule.code, rule.category, category.code, structure.code",
         default=False,
     )
+    worked_days_from_prev_month = fields.Boolean(
+        config_parameter="payroll.worked_days_from_prev_month",
+        string="Fetch worked days and leaves from the previous month",
+        help="Pay fixed items for this month, with adjustments for previous month",
+        default=False,
```
mtelahun: I don't understand what this means. For example, if WORK100 for this month is 20 days and also 20 days for the previous month will the WORK100 payslip input line have 40 days? Also, the help message doesn't make this any clearer. Please consider re-wording this to make it clearer what the purpose of this setting is.

payroll/models/hr_payslip.py
```diff
@@ -308,6 +308,7 @@ def get_worked_day_lines(self, contracts, date_from, date_to):
         applied for the given contract between date_from and date_to
         """
         res = []
+        date_from, date_to = self._get_param_prev_month(date_from, date_to)
```
mtelahun: Please help me understand what's happening here. If `worked_days_from_prev_month` is set then  `date_from` and `date_to` get reset to the previous month's dates? When are the current month's dates evaluated then?

payroll/models/res_config_settings.py
norlinhenrik: The contract has defined fixed salary per calendar month. We pay in the middle of the month. When an employee takes unpaid leave, we register the unpaid days so that the employee will get less salary. Until now we have done this manually on the payslip. But we would like to do it automatically and get the information from the Time Off. The problem is that the Time Off registration for a month may not be complete in the middle of the month. Therefore we would like to deduct unpaid leave with one month delay. When the new option is active, the payslip area WORKED DAYS will show worked days & leaves from the previous month. I am not using `hr_payroll_period`. More development would be needed to be compatible with this module, and then the option should be about "previous period" instead of "previous month".

norlinhenrik: So if the option is active, WORK100 will show days worked in the previous month only.

payroll/models/hr_payslip.py
norlinhenrik: The option will create a system parameter `payroll.worked_days_from_prev_month`. If the value is `True`, then `date_from, date_to = self._get_param_prev_month(date_from, date_to)` will replace the dates with the beginning & end of the previous month.

nimarosa:
Hello Henrik, i think this can be on another module. 
I don't see why someone would need to fetch last month leaves. 
Maybe in another module will be okay if anyone want to use it. Now the payroll module have all the hooks to do that in external modules so i think we can do it that way. 

Also if this conflicts with payroll_period it's another good reason to be another module, since a lot of people use payroll_period. Finally, i think changing the dates "on the fly" can lead to bugs in another modules depending on payroll too. 

I see that you are changing the maintainer tag, feel free to do it in another PR and i will merge it.

norlinhenrik:
I put the logic into a new module and added a hook in hr.payslip.
@mtelahun @nimarosa What do you think?

mtelahun:
Looks good to me.

PR #145, opened 2024-02-15
[WIP][16.0] imp payroll account group entries

cvinh:
In some country (ie FR or PF), the experts do not want to have separate account moves for payslips.
They want to have a global account move without the employees salary details
@Ana0m


nimarosa:
Hello are you going to finish this PR? Tell me so I remove the stale tag.

cvinh:
> Hello are you going to finish this PR? Tell me so I remove the stale tag.

Yes, please remove stale tag

hitrosol:
I think grouping the journal entries can be made if the payslip_run_id is True by creating the payslip batch. And the journal entries will be created if **all** the payslip have been confirmed by user. 
By this case, we don't have to create field action_group_payslips. We are going to create PR for this case in V14 soon.

cvinh:
> I think grouping the journal entries can be made if the payslip_run_id is True by creating the payslip batch. And the journal entries will be created if **all** the payslip have been confirmed by user. By this case, we don't have to create field action_group_payslips. We are going to create PR for this case in V14 soon.

Thanks for your share. We prefer to keep the option for the company if they want to group the journal entries or not. This is because the payment might be global or individual... with a setup, it's more flexible

PR #146, opened 2024-02-26
[15.0][IMP] payroll_account: allow other modules to influence the values of the journal items

AaronHForgeFlow:
Simplify action_payslip_done by adding prepare methods. This allows third modules to introduce changes in the journal items. Introducing changes may be necessary in case the third module adds a mandatory field that is not filled in, for example, the account_operating_unit module adds the operating unit as mandatory in the journal item.

cc @ForgeFlow

### DIFF
--- a/payroll_account/models/hr_payroll_account.py
+++ b/payroll_account/models/hr_payroll_account.py
@@ -111,159 +111,69 @@ class HrPayslip(models.Model):
                     continue
                 debit_account_id = line.salary_rule_id.account_debit.id
                 credit_account_id = line.salary_rule_id.account_credit.id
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
pre-commit will pass when https://github.com/OCA/payroll/pull/147 is merged

MiquelRForgeFlow (approved):
LGTM

nimarosa (approved):
Changes looks good. Functionality tested in runboat and seems okay. 

@AaronHForgeFlow Can you port this PR when merged to v16.0 too? That way we maintain this improvement in future versions.

PR #147, opened 2024-02-26
[15.0][IMP] update copier from template + fixes to pass pre-commit in payroll module

AaronHForgeFlow:


AaronHForgeFlow:
@nimarosa do you think this is fine?

nimarosa:
Hello @AaronHForgeFlow. Here you are updating the copier template but also changing files in payroll module. 
The PR updating the copier template should be unique and do not change other modules. 

This is to maintain order and traceability of changes in the modules. 

If you want to update copier template please do only this in this PR. Then the t-esc-options to t-options change, should be in another PR.

AaronHForgeFlow:
Hi @nimarosa, I was forced to do so because pre-commit was failing after the update. When that happens I think we can fix the CI in a separate commit by module. For example:  https://github.com/OCA/stock-logistics-warehouse/pull/1877/commits

nimarosa:
Fair enough. You are right. 

/ocabot merge nobump

AaronHForgeFlow:
@nimarosa  Thank you!

PR #150, opened 2024-04-11
[FIX] payroll: properly show fields on contract

kbapt:
The payroll fields on the contract view were missing their labels.

**Before:**
image

**After:**
image


mtelahun (approved):
This allways used to bug me, but I never got around to fixing it. Thanks for doing it :grin:

norlinhenrik (approved):

Issue #170, opened 2024-10-23
hr_payroll module for l10n_mx_payroll


EdgarRetes:
Hi everyone.

We have a module named `l10n_mx_hr_payroll` for mexican payroll, but it depends on the module `hr_payroll`, which is **Enterprise**. We got the authorization to open source it but we would like to keep the compatibility with Enterprise. 

Our proposed solution is to create a module named `hr_payroll` in this repository that would consist of just a manifest having a dependency on the `payroll` OCA module. This way, our module would be compatible with the Community and Enterprise versions. 

Any other option? What do you think?

Thanks :)

EdgarRetes:
@max3903

dantrix:
Hola edgard, vi tu comentario. si ustedes son partners puedes implementar EE en tus proyectos con el respectivo modulo  hr_payroll. Si quieres tener compatibilidad con CE tendrias que usar modulos comunitarios, tuyos o de terceros, en el caso de payroll CE hay 2 buenos modulos de amplia experiencia, la de odoo mates y la cybrosys, suerte

PR #171, opened 2024-11-02
[ADD] hr_payroll

max3903:
#170


nimarosa:
We can't add such module because it will conflict with Enterprise modules, that's why the modules don't have the same name.

max3903:
@nimarosa So what are you suggesting? Maintaining 2 separate modules: one for Community, one for Enterprise?

@OCA/human-resources-maintainers Any other options or opinion?

I believe this question will come back with other localizations. Same answer?

nimarosa:
@max3903 What you can do is to create a module in your localization repository, which will install OCA payroll as hr_payroll. 

We can't have it in the main repository because it leads to confusion and incompatibility issues with enterprise version. We changed the name to allow people using enterprise to install OCA Payroll module if they want. So if we include a module named hr_payroll in the repository, we will have compatibility issues. 

What you need to do is to publish your payroll localization in the Mexican OCA localization repository. That way, you can create the module and maintain compatibility without affecting the payroll repository. 

All l10n localizations belongs to each localization repository, it cannot be published in the main payroll repository. 

You need to do it here: 
https://github.com/OCA/l10n-mexico

I hope I clarified things for you. Regards.

max3903:
@nimarosa So you are suggesting that each and every localization host their own `hr_payroll` module in their own repo, right?

nimarosa:
@max3903 not the full payroll module. I'm saying that they host the hr_payroll dummy module which just install OCA payroll module. 

It's the same you suggested but in a localization repo level and not in the main repository. That if you want to use the same module for enterprise and community. The other solution is to just have a community version and change "hr_payroll" to "payroll" in your manifest.

max3903:
@nimarosa Thanks for your clarification.

PR #172, opened 2024-11-11
[17.0][FIX] payroll: Remove migrations folder from 16.0

victoralmau:
Remove migrations folder from 16.0

@Tecnativa


pedrobaeza (approved):
Of course!

omalbastin:
why these modules are not being added to odoo apps store?

PR #159, opened 2024-05-28
[17.0][FW] payroll: ported 111, 117, 119, 151

norlinhenrik:
Port of #111, #117, #119, #151 from 16.0 to 17.0

PR #180, opened 2025-01-05
[17.0][FW] payroll: ported 111, 117, 119, 151

dreispt:
Supersedes #159

### DIFF
--- a/payroll/README.rst
+++ b/payroll/README.rst
@@ -60,11 +60,11 @@ Authors
 Contributors
 ------------
 
-- Odoo SA <info@odoo.com>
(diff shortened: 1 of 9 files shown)


nimarosa:
Merging. 

@dreispt can you port this to 18 too? So we work in an updated module. 

I think there are other PRs in 14 that need to be ported too. 

Thanks . 

/ocabot merge nobump

norlinhenrik:
@nimarosa I don't think there are other PRs in 14.0 that need to be ported too, because I carefully went through all the PRs to do a complete port. Great that this is now merged into 17.0.

PR #182, opened 2025-01-12
[16.0][IMP] payroll: Show payroll inputs menu to allow faster input edit

giarve:
In config menu.


dreispt (approved):

marcelsavegnago (approved):

nimarosa (approved):

giarve:
Taking advantage of the fact that the  bot did not merge yet, I found that the payroll input is not multi-company compatible (it will show all payslip input records from another company).

If anybody wants to suggest changes to fix this, please go ahead or merge it but take into account that it may leak data from another company 

EDIT: actually hr.payslip.run has the same problen in 16.0

PR #184, opened 2025-01-26
[18.0][MIG] hr_payroll_period: migrate to 18.0

dreispt:
- [14.0] [ADD] hr_payroll_period
- [MIG] hr_payroll_period: migration to 15.0
- [FIX] hr_perioll_period: fields definition
- Translated using Weblate (Spanish)
- [IMP] hr_payroll_period: Add a cron to create next fiscal year automatically
- [MIG] hr_payroll_period: Migration to 16.0
- [UPD] Update hr_payroll_period.pot
- [BOT] post-merge updates
- [IMP] hr_payroll_period: pre-commit execution
- [MIG] hr_payroll_period: Migration to 18.0
- [MIG] payroll,hr_payroll_period: changes from migration to 17.0
- [MIG] hr_payroll_period: migration to 17.0


dreispt:
NOTE: after working on it, my feeling is that this module should be discontinued. So I moved this to Draft, as it is not my intention to have it merged.

nimarosa:
@dreispt why do you think we should discontinue this module? I think is useful for people who have different payment schedules or periods and reduce the error selecting them manually. But I'm curious of why you think it's not worth to merge, maybe you have a better way to offer that functionality?

dreispt:
I'm not against adding the concept, I dislike the implementation; I see a lot of complexity.
The date_range dependency seems overpowered.
And I would prefer to have an LGPL license.

nimarosa:
@dreispt finally discovered you are right. The implementation it's not good and i think it's not a valuable module for 18.0
Maybe we can think of a better implementation. Closing again.

PR #188, opened 2025-02-12
[18.0][REF] payroll:  use recordsets instead int lists

dreispt:
<s>I see there was a fix at https://github.com/OCA/payroll/pull/83, but in my tests this was not working.
I improved the tests for the child rule calculation dependencies, and now they are passing.</s>

### DIFF
--- a/payroll/models/hr_contract.py
+++ b/payroll/models/hr_contract.py
@@ -38,8 +38,5 @@ class HrContract(models.Model):
                  hierachy (parent=False first, then first level children and
                  so on) and without duplicates
         """
-        structures = self.mapped("struct_id")
-        if not structures:
-            return []
-        # YTI TODO return browse records
-        return list(set(structures._get_parent_structure().ids))
+        # TODO: remove, too simple and not used
+        return self.struct_id.get_structure_with_parents()
--- a/payroll/models/hr_payroll_structure.py
+++ b/payroll/models/hr_payroll_structure.py
@@ -70,16 +70,12 @@ class HrPayrollStructure(models.Model):
 
     def get_all_rules(self):
         """
-        @return: returns a list of tuple (id, sequence) of rules that are maybe
-                 to apply
+        @return: recordset with all struct rules, ordered by sequence
         """
-        all_rules = []
-        for struct in self:
-            all_rules += struct.rule_ids._recursive_search_of_rules()
-        return all_rules
+        return self.rule_ids._recursive_search_of_rules().sorted("sequence")
 
-    def _get_parent_structure(self):
-        parent = self.mapped("parent_id")
-        if parent:
-            parent = parent._get_parent_structure()
-        return parent + self
+    def get_structure_with_parents(self):
+        if not self:
+            return self.env["hr.payroll.structure"]
+        else:
+            return self.parent_id.get_structure_with_parents() + self
--- a/payroll/models/hr_payslip.py
+++ b/payroll/models/hr_payslip.py
@@ -1,4 +1,4 @@
-# Part of Odoo. See LICENSE file for full copyright and licensing details.
+# Part of Odoo.  LICENSE file for full copyright and licensing details.
 
 import logging
 import math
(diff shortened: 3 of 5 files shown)


nimarosa:
Hello @dreispt i did a fast review of the code. I think you misunderstood the fix in https://github.com/OCA/payroll/pull/83

Child rules can be calculated before and after the parent rule, this is not a limitation and my PR never intended to modify their calculation order. 

The fix i made it's intended to discard the calculation of child rules if the condition of the parent rule is False. In other words, child rules should not be calculated if the parent rule condition is false (so it won't be shown in the payslip). 

In normal use, child rules contains auxiliary calculations that can or are related to the parent rule. So if the condition in parent rule don't returns True, it's not necessary to even calculate the child rules, so they are discarded.

That don't have anything to do with the order of calculation which is what i think you are proposing here. Rules (child or not) should be calculated according their sequence and that is the only field that decides the order of calculation. For example, you can have a parent rule with 2 child rules and one can be calculated before and other later, it depends of the sequence and that don't change with 83 PR.

dreispt:
@nimarosa Understood - if parent rule is not "activated" then there is no point in activating the child rule.

So my PR is fixing something different.
It is a valid effort, correct?

nimarosa:
Hello @dreispt, 

The problem I think with this patch/fix is that you are changing the way payroll historically works. 

Why a child rule should be calculated before the parent rule? I think rules should continue to be calculated according to their sequence. 

I saw cases where child rules need to be calculated after the parent rule, an example of this is: if you need the value of the parent rule to use in the formula of the child rule. 

So if we merge this, we might be broking some payroll implementations and localizations. 

I personally don't have a case when I use child rules after the parent, because I use them as aux variables for the parent rule, but I seen cases in my country which use them after parent. 

I think we don't have to mess with the sequence order. 

The sequence field is there for a reason: to let the user choose the order of calculation. If we change that for child rules, we might affect users. 

Maybe I'm wrong I would like to see what other users think. 
It doesn't affect me so I tend to agree with the majority here. But I think is a sensitive change.

dreispt:
@nimarosa updated, to honor strict Rule sequence. Tests were not changed.

payroll/models/hr_contract.py
```diff
@@ -38,8 +38,5 @@ def get_all_structures(self):
                  hierachy (parent=False first, then first level children and
                  so on) and without duplicates
         """
-        structures = self.mapped("struct_id")
-        if not structures:
-            return []
-        # YTI TODO return browse records
-        return list(set(structures._get_parent_structure().ids))
+        # TODO: remove, too simple and not used
```
nimarosa, line 41: This TODO is asking to remove the line you just changed or  im understanding it wrong?

nimarosa (commented):
I think the changes look right. I didn't have time to test it in runboat. 

So to be clear, this is not affecting calculation order and we still doing it by sequence?

The other changes that use recordsets instead of structs look right to me. 

I think you tested if it's not impacting in calculation times in a noticeable way. The reason to use structs is for speed in payslips with a lot of rules to be calculated, but I think now the speed with be the same using recordsets than in prior versions of Odoo. 

In case we merge this will be good to port it to 17 too.

payroll/models/hr_contract.py
dreispt: This method can be removed, as it implements so useful logic at all.
But I hesitated removing it, for backward compatibility.

dreispt:
@nimarosa The rule sequence is honored when processing payslips. This is ensured [here](https://github.com/OCA/payroll/pull/188/files#diff-21e3fd40769d433ecbbeafce28e3a74b7862b8d1f2740e8a68299ab4ae9e83bfR75).

High calculation efficiency is on my list of important things to improve.
Recordsets are more efficient than record ids - they have a cache and avoid inefficient id-to-record-to-id dances.
In addition they make the code simpler and smaller. This small PR was able to removes almost 30 LoC.

dreispt:
Any blocker for this merge.
This is on step forward to refactor towards new API (as in v8.0) coding practices.

nimarosa:
@dreispt It's okay to merge. Please rebase to be sure every test it's green. I think runboat it's struck.

nimarosa (approved):

dreispt:
Rebased

nimarosa:
@dreispt check that tests are not passing. We need to fix that in order to merge.

dreispt:
@nimarosa Added a commit fixing the payroll tests. It seems to be a preexisting issue, not related to these changes.

PR #198, opened 2025-05-05
[18.0][FIX] payroll_account: Fix key error which arrises 

nimarosa:
We have a key error when trying to create a payslip .this fix the error.


nimarosa:
Superseeded by #199

PR #199, opened 2025-05-05
[18.0][FIX] Fix leaves code population and fix view in salary rules

nimarosa:
Some fixes: 
- Fixed an error when creating a payslip without defined ledger. 
- Since we are now using work_entry model to calculate leaves, the previous code was not fetching the correct codes, it was using GLOBAL code for everything. Now with this fix, we fetch the code from work_entry_type. 
- Also we are using the work_entry_type sequence and not a fixed one. Also we removed file "hr_leave_type" which is not longer used or required. 
- I fixed salary_rule view, was showing "quantity" field in the "code" calculation option. That's not right since it's not used in code calculations. Also quantity was not showing in "fix" and "percentage" types.


nimarosa:
@norlinhenrik @dreispt
Do you want to review so i can merge? Since you are one of the most involved in this repository. You might be interested in this changes. 

This is to make payroll module to use work_entries instead of creating a duplicated field to get the leaves "code". That's the way it works in enterprise so i think we might use work_entries to fetch the leaves and their codes since odoo removed from version 16.0 the "code" field in hr_leave_type and moved it to work_entries. In version 16.0 we recreated the field which i'm removing here, but it's not correct because base odoo it's now using work_entries code which makes more sense. 

Also i fixed some other bugs here. 

My next PR will be creating all the views so we can visualize the work_entries like in enterprise, so we can work with them. Today, community supports work_entries but it has not views for them.

mtelahun (approved):
Looks good.

nimarosa:
Merging since it's necessary to fix the bug and allow to use the module... 

/ocabot merge patch

PR #200, opened 2025-05-19
Employee payslips

vishnumathi-hub:
Employee payslips


dreispt (commented):
I feel some explanation is needed on the proposed changes and reason behind them.
What is "IFSC Code", is it localization specific?

nimarosa:
For me there is no reason to modify the reports since every localization will need different reports for payslips. 

I also suggest it can be a good idea to remove the report entirely and only leave the module with a basic implementation for printed payslips reports, and leave every country or partner to develop his own report. 

We never going to agree on a report format for payroll.

PR #189, opened 2025-02-25
[MIG 17.0] payroll_account

marionumza:

PR #205, opened 2025-07-29
[17.0][MIG] payroll_account

edescalona:
@BinhexTeam

Reasons why the PR was created:

1. There is a #189   for this module, but it has not been continued.
2. The module is merged into the 17.0 branch, but its content does not indicate that it is a valid module, if there is any reason for this, please let me know.
3. Backport 18.0

Hi @dreispt @nimarosa  , if you could review the PR or clarify the issue with the merged module in case there's a reason I'm not aware of, thank you.


payroll_account/migrations/17.0.1.0.1/post-migration.py
rrebollo, line 1: If you don’t have a strong reason to keep it, you should remove the **migrations** folder as part of the standard migration process, as mentioned in the [OCA Migration to version 17.0 guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#tasks-to-do-in-the-migration).

rrebollo (commented):
Thank you for your contribution. This module contains deep functional knowledge, so it’s better to wait for more highly ranked reviews, but I can already see good work here. Feel free to address my comment on migrations scripts.

dreispt:
@edescalona #189 was not a viable PR.
I worked on the 18.0 migration, where relevant cleanup was done. 
I suggest trying the backport of the 18.0 module to 17.0.

dreispt:
PS: the 18.0 migration was based on 16.0 of course, so you might want to look at the additional commits done for 18.0 and consider cherry-picking them.

edescalona:
Hi @dreispt , if I migrated from 18.0, all that's missing is a review of this PR, both functional and code-wise. Thanks for your feedback.

dreispt (approved):
LGTM

nimarosa (approved):

nimarosa:
Just reviewed it via runboat. It looks like it works okay. 
Will merge so we have the module in 17 and then we can make fixes if problems arrises, but look ok for now. Pending testing production. 

/ocabot merge nobump

PR #206, opened 2025-07-30
[IMP] payroll: hr_payslip rule tools add float_compare and relativedelta

giarve:
Missing functions in tools dict

float_compare I am not sure yet if it is necessary (only for complex calcs), but relativedelta improves readability a lot and is already available in the code, just need to expose it in the dict


nimarosa:
Yes there are useful tools. 

/ocabot merge minor

PR #158, opened 2024-05-28
[16.0][FW] payroll_account: Fix multicompanies and refactoring (port of #121 and #123)

norlinhenrik:
Port of #121 and #123 from 15.0 to 16.0

PR #210, opened 2025-09-16
[IMP] payroll_account: use prepare methods in action_payslip_done

jdoutreloux:
Different `_prepare_*` methods were added to this module in #158 but they weren't used in the action_payslip_done method.


jdoutreloux:
@norlinhenrik  as you were the one who did the mentionned PR, could you also review this please?

kaynnan (approved):

PR #217, opened 2025-10-02
[15.0][IMP] hr_payroll_period: add a company parameter to select the months in advance for creating the new HR periods 

AaronHForgeFlow:
Currently users cannot select how much in advance the periods are created.

image

I can put days instead of months...

### DIFF
--- a/hr_payroll_period/models/res_company.py
+++ b/hr_payroll_period/models/res_company.py
@@ -1,0 +1,26 @@
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    payroll_fiscalyear_creation_months_before = fields.Integer(
+        string="Months Before HR Fiscal Year Creation",
+        default=1,
+        help="How many months before the end of the HR fiscal year "
+        "a new one should be created.",
+    )
+
+    @api.constrains("payroll_fiscalyear_creation_months_before")
+    def _check_fiscalyear_creation_months_before(self):
+        for company in self:
+            if (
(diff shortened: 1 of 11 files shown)


hr_payroll_period/models/res_company.py
```diff
@@ -0,0 +1,11 @@
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    fiscalyear_creation_months_before = fields.Integer(
```
CristianoMafraJunior: Improvement suggestion:

Constraint: add a validation to ensure that the `fiscalyear_creation_months_before` field only accepts values between 0 and 12. In practice, it doesn’t make sense to create the next payroll fiscal year more than 12 months in advance, since this may lead to inconsistent records and confusion in reports.

Rename field: consider renaming it to `payroll_fiscalyear_creation_months_before`, making it more explicit that this parameter is related to payroll fiscal years. This avoids ambiguity in case other modules need similar parameters in the future.

Example constraint:

```
@api.constrains("fiscalyear_creation_months_before")
def _check_fiscalyear_creation_months_before(self):
    for company in self:
        if company.fiscalyear_creation_months_before < 0 or company.fiscalyear_creation_months_before > 12:
            raise ValidationError(
                _("Payroll fiscal year creation months before must be between 0 and 12.")
            )

```

AaronHForgeFlow: I agree with that and I changed the code. Thank you!

CristianoMafraJunior (approved):

marcelsavegnago (approved):

kaynnan (approved):

PR #220, opened 2025-10-06
[16.0][IMP] payroll: set currency_id in payslip line

CristianoMafraJunior:
Fields total and amount as Monetary

PR #232, opened 2025-11-07
[16.0][FIX] payroll: Fix warnings for "digits" parameter

anusriNPS:
Warnings generated from "hr.payslip.line" model for "digits" parameter is addressed.
   
   Below is the warning seen:
   2025-11-07 10:01:16,587 21 WARNING devel odoo.fields: Field hr.payslip.line.amount: unknown parameter 'digits', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it

### DIFF
--- a/payroll/models/hr_payslip_line.py
+++ b/payroll/models/hr_payslip_line.py
@@ -47,6 +47,9 @@ class HrPayslipLine(models.Model):
         "Allow editing", compute="_compute_allow_edit_payslip_lines"
     )
 
+    def _valid_field_parameter(self, field, name):
+        return name == "digits" or super()._valid_field_parameter(field, name)
+
     def _compute_allow_edit_payslip_lines(self):
         self.allow_edit_payslip_lines = (
             self.env["ir.config_parameter"]


CristianoMafraJunior (approved):

HekkiMelody:
This fix indeed removes the warning, but I don't think it's correct. Rather, we should simply remove the `digits` parameter as it's not used for fields of type Monetary. (previously the fields were float, and there the `digits` parameter was valid and indeed used by odoo. Now it's no longer used. See #220 )

CristianoMafraJunior:
> This fix indeed removes the warning, but I don't think it's correct. Rather, we should simply remove the `digits` parameter as it's not used for fields of type Monetary. (previously the fields were float, and there the `digits` parameter was valid and indeed used by odoo. Now it's no longer used. See #220 )



> This fix indeed removes the warning, but I don't think it's correct. Rather, we should simply remove the `digits` parameter as it's not used for fields of type Monetary. (previously the fields were float, and there the `digits` parameter was valid and indeed used by odoo. Now it's no longer used. See #220 )

the fix has been made here #234

kaynnan (approved):

anusriNPS:
> This fix indeed removes the warning, but I don't think it's correct. Rather, we should simply remove the `digits` parameter as it's not used for fields of type Monetary. (previously the fields were float, and there the `digits` parameter was valid and indeed used by odoo. Now it's no longer used. See #220 )

Thank you for the input @aleuffre, I did not realize the fields change from "float" to "monetary". As, new fix of removing "digits" parameter handled #234 , closing this PR.

PR #236, opened 2025-12-17
[16.0][ADD] hr_payroll_document_pymupdf: Try harder to read PDF payslip

SirPyTech:
This change allows to read PDFs that couldn't be read by PyPDF, trying to read them again with PyMuPDF.

I also included some refactoring to allow easier override of a few pieces of code, like employee's identification and PDF parsing.


quirino95 (approved):

hr_payroll_document/wizard/payroll_management_wizard.py
```diff
@@ -22,48 +25,123 @@ class PayrollManagamentWizard(models.TransientModel):
...
+            "name": _("Payrolls sent"),
+            "type": "ir.actions.act_window",
+            "res_model": "hr.employee",
+            "views": [
+                (
+                    self.env.ref("hr.hr_employee_public_view_kanban").id,
+                    "list",
+                )
+            ],
+        }
```
quirino95, line 94: With this `return`, frontend looks "stuck" after sent payrolls.
suggestion:
`return {"type": "ir.actions.act_window_close"}`

Prevents also redirection

hr_payroll_document/wizard/payroll_management_wizard.py
```diff
@@ -22,48 +25,123 @@ class PayrollManagamentWizard(models.TransientModel):
...
+
+        for employee, pages in employee_to_pages.items():
+            path = self._build_employee_payroll(
+                _(
+                    "Payroll %(subject)s %(employee)s.pdf",
+                    employee=employee.name,
+                    subject=self.subject,
+                ),
+                pages,
+                encryption_key=employee.identification_id,
```
quirino95: This means that every pdf will have employee's identification id.
What if we left the choice of whether or not to have a password up to the employees themselves rather than a "common" (easy) one?

hr_payroll_document/wizard/payroll_management_wizard.py
HekkiMelody: To be precise, only the `list` view is available, and not the `form`, so clicking on any employee doesn't actually do anything. That's why it "looks stuck".

The problem already existed before, it's not been introduced by this PR.

SirPyTech: This behavior was already present, please see the other comment about the PR's purpose.

hr_payroll_document/wizard/payroll_management_wizard.py
SirPyTech: This behavior was already present, please see the other comment about the PR's purpose.

SirPyTech (commented):
Thanks for the review!
Indeed this module had a few issues that you have quickly spotted :clap:

This PR only aims to change the behavior as described
> read PDFs that couldn't be read by PyPDF, trying to read them again with PyMuPDF

I think it's better to add other fixes/improvements in separate PRs so that they can be independent.

Please let me know if there is something wrong in the new behavior introduced by this PR.

hr_payroll_document/wizard/payroll_management_wizard.py
```diff
@@ -1,7 +1,10 @@
 import base64
+import io
 from base64 import b64decode
 
-from pypdf import PdfReader, PdfWriter
+import pymupdf
+from pypdf import PdfReader, PdfWriter, errors
+from reportlab.pdfgen import canvas
```
peluko00: What's the purpose to use new libraries and extract the logic in different functionsthat only use one time?

peluko00 (commented):
LGT

hr_payroll_document/wizard/payroll_management_wizard.py
SirPyTech: Hi, thanks for having a look!

> What's the purpose to use new libraries [...]?

The new libraries are being used because some PDF could not be read by the library (PyPDF) that had been used, like the one added in the new test.

> What's the purpose to [...] extract the logic in different functionsthat only use one time?

This allows depending modules to override specific parts of the flow without the need to override the whole flow.
For instance, we have overridden the parsing of one page in a private module because our PDF has a custom structure and benefits from a dedicated implementation.

peluko00: Thanks for the explanation @SirPyTech!

peluko00 (approved):
LGTM! Code review

peluko00:
Can you put on v17 please @SirPyTech ?

peluko00:
Can you merge @nimarosa please?

SirPyTech:
> Sorry @peluko00 you are not allowed to merge.
> 
> To do so you must either have push permissions on the repository, or be a declared maintainer of all modified addons.
> 
> If you wish to adopt an addon and become it's [maintainer](https://odoo-community.org/page/maintainer-role), open a pull request to add your GitHub login to the `maintainers` key of its manifest.

By the way, this is probably because the PR is also editing the `requirements.txt` file that is outside the maintained module.

SirPyTech:
> /ocabot merge minor

> Sorry @peluko00 you are not allowed to merge.
> 
> To do so you must either have push permissions on the repository, or be a declared maintainer of all modified addons.
> 
> If you wish to adopt an addon and become it's [maintainer](https://odoo-community.org/page/maintainer-role), open a pull request to add your GitHub login to the `maintainers` key of its manifest.

Can someone of @OCA/human-resources-maintainers please merge?
The module's maintainer would like to merge but is not allowed to do it.
Thanks!

pedrobaeza (commented):
I personally don't think the dependencies should be changed in a stable module unless hard reasons. In this case, I see that this second try on PDF can be performed in a new module instead of adding the other PDF library here. Maybe adding a hook in the base module is needed, but that's not a problem.

SirPyTech:
> I personally don't think the dependencies should be changed in a stable module unless hard reasons. In this case, I see that this second try on PDF can be performed in a new module instead of adding the other PDF library here. Maybe adding a hook in the base module is needed, but that's not a problem.

Thanks for having a look!
Right, they shouldn't change in a **Stable** module, but the one this PR is changing is still in **Beta**: https://github.com/OCA/payroll/blob/a49680ae17ea75e6710fe3ea27bbd29818238f12/hr_payroll_document/README.rst?plain=1#L13-L15

pedrobaeza:
For me, stable is not that development status that nobody cares and changes it, and that's why it has it, as it's the default value. The stable status is that this version (16.0) is 3 years old and the module is here since a lot (almost 2 years). Changing dependencies will break all the existing installations having it.

SirPyTech:
We should be more careful about the development status.

Ok I'll split the "harder reading" in another module and set the dev status of `hr_payroll_document` to stable so we're all happy :smile:

pedrobaeza:
OK, thanks for considering. I still think the stability for these things is not defined by that development status, as strictly this may not comply with the status rules, but the time the module landed in OCA, so doing breaking changes may affect a lot of users. That status is more like a "current quality status".

SirPyTech:
Added the stable development status in https://github.com/OCA/payroll/pull/243, and the new Python dependency has been moved to a new module.

I have left the `mock_identification` method in `hr_payroll_document`'s common setup for tests because it might be useful for other dependencies too.

Let me know what you think :smile:

pedrobaeza (approved):

CristianoMafraJunior (approved):

peluko00 (approved):
LGTM!

SirPyTech:
Rebased to resolve conflicts

quirino95 (approved):
Code and functional review: LGTM

SirPyTech:
> /ocabot merge minor

> Sorry @peluko00 you are not allowed to merge.
> 
> To do so you must either have push permissions on the repository, or be a declared maintainer of all modified addons.
> 
> If you wish to adopt an addon and become it's [maintainer](https://odoo-community.org/page/maintainer-role), open a pull request to add your GitHub login to the `maintainers` key of its manifest.

Can someone of @OCA/human-resources-maintainers please merge?
The module's maintainer would like to merge but is not allowed to do it.
Thanks!

(copy of comment https://github.com/OCA/payroll/pull/236#issuecomment-3699315787)

PR #245, opened 2025-12-31
[IMP] hr_payroll_document: Optional payroll encryption

SirPyTech:
Improvement proposed in https://github.com/OCA/payroll/pull/236#discussion_r2635484463.

Allow the employee or HR to disable encryption of the employee's payroll.

### DIFF
--- a/hr_payroll_document/models/hr_employee.py
+++ b/hr_payroll_document/models/hr_employee.py
@@ -1,9 +1,17 @@
-from odoo import _, models
+from odoo import _, fields, models
 from odoo.exceptions import ValidationError
 
 
 class Employee(models.Model):
     _inherit = "hr.employee"
+
+    no_payroll_encryption = fields.Boolean(
+        string="Disable payrolls encryption",
+        help="If this is disabled (default), "
+        "the PDF payrolls are encrypted using the Identification No.\n"
+        "Only future payrolls are affected by this change, "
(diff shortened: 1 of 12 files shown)


hr_payroll_document/models/hr_employee.py
```diff
@@ -1,10 +1,18 @@
-from odoo import _, models
+from odoo import _, fields, models
 from odoo.exceptions import ValidationError
 
 
 class Employee(models.Model):
     _inherit = "hr.employee"
 
+    no_payroll_encryption = fields.Boolean(
```
peluko00, line 8: By default please encrypt de document

peluko00 (changes requested):

hr_payroll_document/models/hr_employee.py
FernandoRomera: Please do not use negative Booleans.
Change to "payroll_encryption"

SirPyTech: Hi, yes I had the same thought, that's why the `Boolean` is "Do **not** encrypt the document": by default `Boolean`s are `False`, so the document is encrypted.

peluko00 (approved):
LGTM!

/ocabot merge minor

peluko00:
Can you forward to major versions please @SirPyTech?

PR #248, opened 2026-01-05
[18.0][FIX] payroll: make again payslip name editable at draft state

cgarcia-solvos:
Fw-port of https://github.com/OCA/payroll/pull/242


pedrobaeza (approved):
Fast-tracking as fw-port:

/ocabot merge patch

PR #252, opened 2026-02-11
[ADD] hr_payroll_document_queue

SirPyTech:
Add a new module to process the payslips asynchronously.

I have also included a small improvement to allow multiple wizards to be processed at the same time without conflicts.
A more elaborate solution could be implemented (like using https://docs.python.org/3/library/tempfile.html) but that would require more refactoring.

Let me know what you think!


quirino95 (approved):
Code and functional review: LGTM

**Suggestion (non-blocking)**: What do you think about keeping just the "Send" button instead of two? Of course, in this case, "Send" will do an asynchronous call. I think this is better for the user experience.

hr_payroll_document_queue/wizards/payroll_management_wizard_views.xml
```diff
@@ -0,0 +1,59 @@
...
+                <attribute name="attrs">{
+                    'invisible': [
+                        ('is_send_visible', '=', False),
+                    ],
+                }</attribute>
+            </button>
+            <button name="send_payrolls" position="after">
+                <button
+                    string="Send Async"
+                    name="send_payrolls_async"
```
anusriNPS, line 47: **Suggestion (non-blocking):**  When there is no attachment selected, user able to execute send/send Async. We can enable these buttons only when there is an attachment or raise `UserError` when trying to send without any attachment

anusriNPS (approved):
Code and Functional Review: LGTM.

**Observation:** After emails sent, could not check notifications in Odoo chatter box about emails.

hr_payroll_document_queue/wizards/payroll_management_wizard_views.xml
SirPyTech: This could be a nice improvement for the `hr_payroll_document` module, but since it works this way I prefer to keep the new _async flow_ as close as possible to the existing _sync flow_.

SirPyTech:
Thanks both for your reviews!

> **Suggestion (non-blocking)**: What do you think about keeping just the "Send" button instead of two? Of course, in this case, "Send" will do an asynchronous call. I think this is better for the user experience.

@quirino95 I prefer to show explicitly what's going to happen, one button or the other can be hidden for specific users adding user groups.

---
> **Observation:** After emails sent, could not check notifications in Odoo chatter box about emails.

@anusriNPS we already talked about this and that was because jobs weren't starting, is this still happening?

anusriNPS:
> > **Observation:** After emails sent, could not check notifications in Odoo chatter box about emails.
> 
> @anusriNPS we already talked about this and that was because jobs weren't starting, is this still happening?

I was able to verify email notification from `mailhog` but I thought it would be good to notify in Odoo chatter box about mail sent information for user of payroll management or provide some information about notification related to employee details and so on. Currently, only notification says `Payroll sent successfully` which is not intuitive. This observation is also related to `hr_payroll_document_queue` in general not specific to this PR.

hr_payroll_document_queue/wizards/payroll_management_wizard_views.xml
anusriNPS: Agree, keeping both buttons flow similar is fine.

SirPyTech:
> > > **Observation:** After emails sent, could not check notifications in Odoo chatter box about emails.
> > 
> > 
> > @anusriNPS we already talked about this and that was because jobs weren't starting, is this still happening?
> 
> I was able to verify email notification from `mailhog` but I thought it would be good to notify in Odoo chatter box about mail sent information for user of payroll management or provide some information about notification related to employee details and so on. Currently, only notification says `Payroll sent successfully` which is not intuitive. This observation is also related to `hr_payroll_document_queue` in general not specific to this PR.

I think we are mixing up the 2 types of notifications that the user can select: image
If the user selects "Handle by Emails", they will only get an email; if they select "Handle in Odoo", they will only get a notification in Odoo.

If they want Odoo notifications, then in the chatter there will be a message with the same subject of the sent payrolls image
and in Discuss they'll be able to also open the wizard's record: image, and then image.

I agree that it could be improved but if you have a look at the code, I simply tell Odoo to notify the user with `self.env["mail.thread"].message_notify` and then everything follows the standard flow.

anusriNPS:
> > > > **Observation:** After emails sent, could not check notifications in Odoo chatter box about emails.
> > > 
> > > 
> > > @anusriNPS we already talked about this and that was because jobs weren't starting, is this still happening?
> > 
> > 
> > I was able to verify email notification from `mailhog` but I thought it would be good to notify in Odoo chatter box about mail sent information for user of payroll management or provide some information about notification related to employee details and so on. Currently, only notification says `Payroll sent successfully` which is not intuitive. This observation is also related to `hr_payroll_document_queue` in general not specific to this PR.
> 
> I think we are mixing up the 2 types of notifications that the user can select: image If the user selects "Handle by Emails", they will only get an email; if they select "Handle in Odoo", they will only get a notification in Odoo.
> 
> If they want Odoo notifications, then in the chatter there will be a message with the same subject of the sent payrolls image and in Discuss they'll be able to also open the wizard's record: image, and then image.
> 
> I agree that it could be improved but if you have a look at the code, I simply tell Odoo to notify the user with `self.env["mail.thread"].message_notify` and then everything follows the standard flow.

Thank you for the detailed response.  It helped to understand different notifications sent as per configuration. Agree, we are following standard flow here. Discussion was helpful to understand the module better.

SirPyTech:
@OCA/payroll-maintainers this has been ready for merge for 4 months, can someone please have a look?
@peluko00 here I did a small improvement to a module you are maintaining, could you please have a look?
Thanks!

nimarosa (approved):

nimarosa:
Looks good. Only code review. 
/ocabot merge nobump

SirPyTech:
> Looks good. Only code review.
> /ocabot merge nobump

Thanks @nimarosa for trying to merge! Can we try again today?

CristianoMafraJunior (approved):

SirPyTech:
> Looks good. Only code review.
> /ocabot merge nobump

Thanks @nimarosa for trying to merge! Can we try again today?
cc @OCA/payroll-maintainers

PR #259, opened 2026-04-09
[FIX] hr_payroll: validate computed rule numeric values

JuanMarta:
This change adds a type check in `_compute_rule` to ensure that
the computed numeric values are returned as floats.

It validates and normalizes the following keys:
- `quantity`
- `rate`
- `amount`

A `UserError` is raised when the returned value is invalid.

### DIFF
--- a/payroll/models/hr_salary_rule.py
+++ b/payroll/models/hr_salary_rule.py
@@ -199,7 +199,6 @@ class HrSalaryRule(models.Model):
         self.require_code_and_category = require
         return require
 
-    # TODO should add some checks on the type of result (should be float)
     def _compute_rule(self, localdict):
         """
         :param localdict: dictionary containing the environement in which to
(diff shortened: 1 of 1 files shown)


nimarosa (approved):

nimarosa:
@JuanMarta

The change it's okay but types are checked on field level. Since fields are float and integer type, this change is redundant. 

We can merge it if you consider this useful in some way but probably we will never raise this error. 

Also please check pre commit and tests if you want the PRs to be merged.

JuanMarta:
Thanks for the review.

I added this because there was an explicit TODO in `_compute_rule` saying that some checks should be added on the type of the result.

My intention was mainly to address that TODO and make the expected contract of the returned values a bit more explicit at this level.

That said, your point makes sense: since those values are already handled at field level and this error would likely never be raised in practice, I agree the change may be redundant.

If you prefer, I can close the PR or rework it.

dreispt (approved):

nimarosa (approved):

PR #261, opened 2026-05-14
[18.0][IMP] payroll: add opening values (seniority and leave balances) for …

cvinh:
…contract - helped by Claude Opus-4.7


cvinh:
commit message updated to fit AI_POLICY

PR #270, opened 2026-05-21
[FIX] hr_payroll_document: add .pdf extension to payroll attachment name

fmontesdev:
## Problem

  Payroll attachments were sent without a `.pdf` file extension.
  On desktop operating systems, the file is opened correctly because
  the MIME type is inferred from the file content. However, on mobile
  devices (iOS and Android), email clients rely on the file extension
  to determine how to open the attachment, making the payroll
  unreadable on mobile.

  ## Solution

  Add `.pdf` extension to the attachment name when creating the
  `ir.attachment` record in `send_mail`.

  ## Tests

  Added `test_attachment_name_has_pdf_extension` to verify that
  the generated attachment name ends with `.pdf`.


peluko00 (approved):
LGTM!

peluko00:
Could please @fmontesdev backport to future versions?

PR #266, opened 2026-05-18
[18.0][IMP] payroll_contract_advantages: add complex mechanism for advantag…

cvinh:
…e value calculations - helped by Claude Opus-4.7
We added demo data as well

PR #271, opened 2026-05-22
[18.0][ADD] payroll_contract_advantages_base and payroll_contract_advantages_structure : Add abstract module to be able to link

cvinh:
payroll_contract_advantages_template to another object
in another commit, we use that abstract module to link templates with payroll structures


cvinh:
depends on https://github.com/OCA/payroll/pull/266 and https://github.com/OCA/payroll/pull/267

Issue #275, opened 2026-07-08
Remove "Company Required" in hr_salary_rull


stferraro:
# Remove `required=True` from `company_id` in Payroll module

## Problem
In the `payroll` module (OCA), the `company_id` field in `hr.salary.rule` and other models is marked as `required=True`. However, in the `payroll_account` module (accounting integration), these same models use `company_dependent=True` instead.

This inconsistency creates unnecessary constraints:
- `required=True` forces every record to have a single company
- `company_dependent=True` is more flexible - it allows field values to vary per company

## Solution
Remove `required=True` from `company_id` fields in the following models and follow the pattern used in `payroll_account`:
- `hr.salary.rule`
- Any other base payroll models with `company_id` as required

This aligns with how the accounting integration layer handles multi-company scenarios.

## Affected Files
- `addons_base/payroll/payroll/models/hr_salary_rule.py`
- Review other payroll models for consistency

## References
- `enterprise/hr_payroll_account/models/hr_salary_rule.py` (uses `company_dependent=True`)

dantrix:
Hi, i think required company is good for multicompany scenarios, +1 for no removed. for example the rule basic for company A is diferent for company B for this reason is required the company field.

raulovallet:
@dantrix > Hi, i think required company is good for multicompany scenarios, +1 for no removed. for example the rule basic for company A is diferent for company B for this reason is required the company field.

That is true in many Odoo scenarios. For example, each payslip belongs to a specific company. However, in this particular case, we are talking about payroll structures. In the Dominican Republic payroll localization, we have more than 34 salary rules. If payroll structures were company-specific, we would have to duplicate all of those rules for every single company, even though the rules are exactly the same across all companies. In addition, whenever there is a fiscal or legal change, we would have to update every duplicated salary rule in every company, significantly increasing maintenance effort and the risk of inconsistencies. By allowing payroll structures to be shared, we would only need to maintain a single payroll structure with one set of salary rules, eliminating unnecessary duplication and making future updates much easier.

A similar example is products. If I create a product called “Pen”, I can create it once and use it across all my companies because the Company field on products is not mandatory. This avoids unnecessary duplication of the same product for each company, and the same principle applies to payroll structures.

dantrix:
Hi Raul, Your point of view is completely valid; you could create a module to modify this for your localization, though I’m not sure if implementing it as a standard feature would be a good idea. IMHO.

raulovallet:
[...]

PR #276, opened 2026-07-08
[19.0][IMP] REMOVE company_id required=True

stferraro:
https://github.com/OCA/payroll/issues/275
-Remove required=True in hr_payroll_structure


nimarosa:
I don't think is ok to remove required=True. A lot of people use multi-company so a structure without company is a orphaned structure usable in all companies. Why is the reason to remove it?

PR #269, opened 2026-05-19
[19.0] [MIG] payroll_account: Migration to 19.0

stferraro:
Related https://github.com/OCA/payroll/pull/260
Related https://github.com/OCA/payroll/pull/228

cc @CristianoMafraJunior  @tishmen @nimarosa @dreispt

PR #282, opened 2026-08-08
[19.0][MIG] payroll_account: Migration to 19.0

neoand:
Migration of `payroll_account` from 18.0 to 19.0.

Module history was replayed with `git format-patch --keep-subject --stdout origin/19.0..origin/18.0 -- payroll_account | git am -3` (69 commits), and the migration itself is the single commit on top.

## What Odoo 19 broke

Odoo 19 removed the `hr_contract` addon: contract data now lives on `hr.version`, which is exactly the model this module extends. The `payroll` module already made that jump on 19.0, so this migration follows the same pattern.

| Change | Fix |
|---|---|
| `hr_contract` addon is gone | `models/hr_contract.py`: `_inherit` `"hr.contract"` → `"hr.version"` |
| Parent view `hr_contract.hr_contract_view_form` no longer exists | Inherit `payroll.hr_contract_view_form` (model `hr.version`) and anchor the accounting fields after `struct_id` |
| `hr.employee.gender` → `sex` | Fixed in tests |
| `hr.employee.bank_account_id` (m2o) → `bank_account_ids` | Fixed in tests |

Tests now build the contract through `employee.version_id.write()` with `contract_date_start`, the pattern used by `payroll/tests/common.py` on 19.0.

The `18.0.1.0.2` migration scripts were dropped.

## Testing

On `odoo:19.0` Community:

```
Module payroll_account: 0 failed, 0 error(s) of 3 tests
```

Beyond the module's own suite, I checked the feature end-to-end on a fresh database: a payslip whose rules carry `account_debit` / `account_credit`, confirmed with `action_payslip_done()`, produces a balanced `account.move` posted on the journal configured on the version.

```
Entry MISC/2026/04/0001 | journal=Miscellaneous Operations | state=posted
ACCOUNT   LABEL          DEBIT       CREDIT
443000    Salary     12,000.00         0.00
201000    Salary          0.00    12,000.00
201000    ISR             0.00     1,200.00
443000    ISR         1,200.00         0.00
          TOTALS     13,200.00    13,200.00
```

I also re-ran a payslip-engine check on `payroll` alone before and after installing this module to confirm nothing regressed.

## Note

Loading demo data changed in Odoo 19 — `--without-demo` is no longer enough and there is a new `odoo module force-demo` subcommand. Without demo data these tests fail on `base.res_partner_12`, which does exist in 19.0's demo. Worth knowing when reproducing locally.

@appstogrow @nimarosa

---

Assisted-by: Claude Opus 4.8


stferraro:
Hi @neoand,

I noticed this PR came in after mine (#269, opened back in May), so I wanted to flag something before they diverge further.

While working on #269 I found and fixed a real bug in payroll_account's journal handling: when generating payslips in bulk from a hr.payslip.run batch, the batch's journal was silently overriding the employee's own contract journal (hr.version.journal_id) instead of respecting it as a fallback. Worth checking if your migration carries the same behavior, since the original v18 code has it too.

Changes in #269 that address this:

create() override in hr.payslip now gives priority to the contract's own journal_id, falling back to the batch's journal only when the contract has none.
Added journal_id to _get_whitelist_fields_from_template() so it's actually copied when a contract template is applied — it wasn't before.
Test coverage for both the standalone create() path and the batch wizard, plus the account-move balancing paths that weren't covered before.
Since #269 predates yours and already has this fix in place with tests, would you mind taking a look and reviewing it? Happy to also look at yours and see if there's anything worth merging between the two. No rush — just wanted to make sure this doesn't get lost with two open migrations for the same module.

Thanks for your work on this either way!
[](https://github.com/OCA/payroll/pull/269)

neoand:
Cierro este PR por decisión interna de pausar contribuciones a OCA a partir del 15-ago-2026.

El código queda archivado en mi fork privado (`neoand/arw-ivan`) para uso local del proyecto de migración ARW.

Agradezco especialmente a @stferraro por avisar a tiempo que existía #269, y por la conversación que se abrió. El trabajo de #269 cubre esta migración con tres aprobaciones, CI en verde, y el fix del bug de `journal_id` en el template de contratos. Cerrar este a favor del suyo era lo correcto.

Gracias a los mantenedores por su tiempo en revisarlo.

— Anderson

PR #284, opened 2026-08-08
[19.0][ADD] payroll_work_entry: compute worked days from work entries

neoand:
New module: build a payslip's worked days from `hr.work.entry`.

## Why

`payroll` derives worked days from the resource calendar — what the employee **should** have worked (`hr_payslip.py`, `_compute_worked_days`, with `compute_leaves=False`). Where an installation records work entries, that data describes what actually **was** worked: attendances, overtime, absences.

For a workforce that clocks in, the difference is not cosmetic. The payslip silently ignores the attendance records the system already has: overtime is never paid and absences are never deducted, because the calendar says every day was a full one.

`hr_work_entry` is Community in 19.0, and `hr.work.entry` is already adapted to `hr.version`, so the bridge is small.

## What it does

- Takes the **validated** work entries falling inside the payslip period
- Groups them by the **payroll code** of their work entry type — one worked-days line per code, so rules can address them as `worked_days.OVERTIME`
- Hours come from the entries; days are derived from the contract calendar's `hours_per_day`, so a 12-hour shift counts as one day rather than one and a half
- A type with no payroll code is skipped: a rule has no way to reference it

**Contracts with no work entry in the period fall back to `payroll`'s calendar computation**, so installing this module changes nothing for them.

## On states

Only `validated` entries count. Draft, cancelled and conflicting ones are deliberately excluded: a conflict is Odoo telling you the record is not trustworthy, and paying it would hide the problem instead of surfacing it.

Worth knowing when adopting this: Odoo validates **one work entry per employee and day**. A second entry on a date that already has a validated one stays in conflict, so overtime recorded on the same day as an attendance needs a work entry model that accounts for it or it will never reach the payslip. That is upstream behaviour, not something this module introduces, but it is the first thing people hit.

## Testing

`0 failed, 0 errors of 7 tests` on `odoo:19.0` Community, on a database **without demo data**.

Coverage: hours taken from the entries rather than the calendar, one line per type, a rule reading the recorded hours, draft/cancelled entries excluded, entries outside the period ignored, days following `hours_per_day`, and the fallback to the calendar when there are no entries.

Fixtures are built in `setUpClass`. Two details that cost me time and are baked into them: a bare `resource.calendar` with no attendance lines makes every entry fall *outside calendar* and unvalidatable — the tests copy the company calendar; and writing `state = 'validated'` directly leaves the record in conflict, so they go through `action_validate()`.

---

Assisted-by: Claude Opus 4.8


CristianoMafraJunior (changes requested):
https://github.com/OCA/.github/blob/master/AI_POLICY.md

neoand:
Thanks for pointing me to the policy — that one's on me, and it's fixed now.

The commit was using `Co-authored-by:` for the AI tool, which the policy rules out. I've amended it
to use the `Assisted-by:` trailer instead and added the disclosure to the description.

Happy to answer anything about the implementation directly — I'd much rather work through review
comments than regenerate the patch. Thanks for taking the time to look at it.

nimarosa (approved):

haivanlive:
In Odoo Community, `hr.work.entry` lacks built-in UI actions/ buttons to validate work entries. Because of this, standard Community users cannot easily move work entries to the `validated` state.

payroll_work_entry/models/hr_payslip.py
```diff
@@ -0,0 +1,78 @@
...
+        ``payroll`` derives worked days from the resource calendar, i.e. what
+        *should* have been worked. Where work entries exist they describe what
+        actually was — attendances, overtime, absences — so they win.
+
+        Contracts with no work entry in the period keep the calendar behaviour,
+        so installing this module changes nothing for those.
+        """
+        self.ensure_one()
+        res = []
+        remaining = self.env["hr.version"]
```
CristianoMafraJunior, line 67: `remaining = self.env["hr.contract"].browse()`  # Empty recordset of the correct model

CristianoMafraJunior (changes requested):

neoand:
Cierro este PR por decisión interna de pausar contribuciones a OCA a partir del 15-ago-2026.

El código queda archivado en mi fork privado (`neoand/arw-ivan`) para uso local del proyecto de migración ARW.

Agradezco a @nimarosa por la aprobación, a @CristianoMafraJunior por la primera revisión (trailer de IA) y por la observación técnica sobre el recordset inicial en `get_worked_day_lines()`, y a @haivanlive por apuntar que la versión Community no trae UI nativa para validar work entries — la discusión sobre validación en Community vale la pena tenerla en OCA cuando se reanude la colaboración.

— Anderson

PR #287, opened 2026-08-17
[18.0][FIX] payroll_contract_advantages: bound validation, data integrity and precision

CristianoMafraJunior:
**What's broken today**

- `hr.contract.advantage._check_bound_limits()` treats `0.0` as a real bound instead of
  "no limit on that side". Any `hr.contract.advantage.template` left with the default
  (unset) `lower_bound`/`upper_bound` therefore rejects **every non-zero amount**,
  making the advantage type unusable until an admin discovers they must also fill in
  bounds that have no `required` flag or hint suggesting so.
- Nothing prevents assigning the **same advantage template twice on the same contract**.
  `payroll_contract_advantages/models/hr_payslip.py` builds a `{code: amount}` dict from
  `contract.advantages_ids`, so a duplicate line silently overwrites one amount with the
  other with no warning, and the resulting payslip total is wrong without any error
  being raised.
- Nothing prevents two `hr.contract.advantage.template` records from sharing the same
  `code`. That code is the key salary rules use to read the value (`advantages.<CODE>`),
  so a duplicate code makes two different advantages collide in the same computation.
- `hr.contract.advantage.contract_id` is not `required` and has no `ondelete`, so
  deleting a contract leaves orphaned `hr.contract.advantage` rows with
  `contract_id = NULL` instead of being cleaned up.
- `hr.contract.advantage.template` allows saving a `default_value` outside its own
  `lower_bound`/`upper_bound`, and allows `lower_bound > upper_bound` — an
  already-inconsistent template that only surfaces the problem later, when someone tries
  to use it.
- `amount` (on the advantage line) and `lower_bound` / `upper_bound` / `default_value`
  (on the template) use plain `Float` with no digits precision, unlike the rest of the
  payroll engine which consistently uses the `Payroll` (2 decimals) precision for
  monetary-like amounts.

**What this PR changes**

- `hr_contract_advantage.py`
  - `_check_bound_limits()`: a bound of `0.0` is now treated as "not set", matching the
    common Odoo convention, instead of blocking every non-zero amount.
  - `contract_id`: now `required=True, ondelete="cascade"`.
  - `advantage_template_id`: now `required=True`.
  - `amount`: now uses `digits="Payroll"`.
  - New SQL constraint `contract_template_unique`: one template per contract.
- `hr_contract_advantage_template.py`
  - New SQL constraint `code_unique`.
  - New `_check_bounds_consistency()`: rejects `lower_bound > upper_bound` and a
    `default_value` outside the template's own bounds (same "`0.0` = no limit"
    convention).
  - `lower_bound` / `upper_bound` / `default_value`: now use `digits="Payroll"`.
- Tests: 5 new test cases covering every rule above, plus one pre-existing test adjusted
  (`test_onchange_advantage_template_sets_default_amount`) whose fixture value became
  inconsistent with the new bound-consistency check.

cc @nimarosa @pedrobaeza


nimarosa (approved):

CristianoMafraJunior:
@nimarosa merge?

payroll_contract_advantages/__manifest__.py
```diff
@@ -15,5 +15,5 @@
         "views/hr_contract_views.xml",
     ],
     "application": True,
-    "maintainers": ["nimarosa"],
+    "maintainers": ["nimarosa, CristianoMafraJunior"],
```
nimarosa: This is one string with a comma in it, not two entries. Should be ["nimarosa", "CristianoMafraJunior"]. You can see what it does in the README diff in this same PR, the avatar URL comes out as github.com/nimarosa, CristianoMafraJunior.png. The OCA bot reads this list too.

payroll_contract_advantages/models/hr_contract_advantage.py
```diff
@@ -8,9 +8,17 @@ class HrContractAdvantage(models.Model):
     _name = "hr.contract.advantage"
     _description = "Employee's Advantages on Contract"
 
-    contract_id = fields.Many2one("hr.contract")
+    _sql_constraints = [
+        (
+            "contract_template_unique",
+            "unique(contract_id, advantage_template_id)",
```
nimarosa: I'd rather this were a Python constraint.

The problem you're describing is real and I'm not arguing with it, a duplicated line does silently clobber one of the amounts in get_current_contract_dict. But unique(contract_id, advantage_template_id) also kills the case where the same advantage is recorded twice with different validity dates, and that's what I'm about to need. I'm working on date_from/date_to on hr.contract.advantage so that recomputing an old payslip gives you the amount that was agreed for that period instead of today's, and consecutive rows for one template are the normal shape of that.

A module inheriting this one can refine an @api.constrains. It can't drop an SQL constraint. So could you move it over?

payroll_contract_advantages/models/hr_contract_advantage_template.py
```diff
@@ -1,18 +1,58 @@
...
 
 
 class HrContractAdvandageTemplate(models.Model):
     _name = "hr.contract.advantage.template"
     _description = "Employee's Advantage on Contract"
 
+    _sql_constraints = [
+        (
+            "code_unique",
+            "unique(code)",
```
nimarosa: There's no migration for this. On a database that already has duplicate codes Postgres refuses the index, Odoo logs a warning and carries on, so the constraint ends up not existing on exactly the databases that needed it. Same story for the NOT NULL on contract_id.

Can you add migrations/18.0.1.1.0/pre-migration.py that clears the orphan rows and does something sensible with duplicate codes? payroll_account has one in this repo you can take the shape from.

payroll_contract_advantages/models/hr_contract_advantage_template.py
```diff
@@ -1,18 +1,58 @@
...
+        )
+    ]
+
     name = fields.Char(required=True)
     code = fields.Char(required=True)
     lower_bound = fields.Float(
-        help="Lower bound authorized by the employer for this advantage"
+        digits="Payroll",
+        help="Lower bound authorized by the employer for this advantage. "
+        "Leave at 0 for no lower bound.",
```
nimarosa: Can you put a line about this in readme/USAGE.md? It changes what an existing install accepts and nothing tells the admin.

payroll_contract_advantages/models/hr_contract_advantage_template.py
```diff
@@ -1,18 +1,58 @@
...
+                raise ValidationError(
+                    _("Lower bound can't be greater than upper bound.")
+                )
+            if template.upper_bound and template.default_value > template.upper_bound:
+                raise ValidationError(
+                    _("Default value can't be greater than upper bound.")
+                )
+            if (
+                template.lower_bound
+                and template.default_value
```
nimarosa: default_value is a value, not a bound, so 0 is a fine thing to want there. This guard means a template with default 0 and lower_bound 10 goes through, while the upper_bound check just above has no equivalent guard. I'd drop it and leave both checks the same shape.

payroll_contract_advantages/README.rst
```diff
@@ -1,7 +1,3 @@
-.. image:: https://odoo-community.org/readme-banner-image
-   :target: https://odoo-community.org/get-involved?utm_source=readme
-   :alt: Odoo Community Association
-
 ===========================
```
nimarosa: Could you drop README.rst and index.html from the PR? They've been regenerated with an old oca-gen-addon-readme, it's removed the OCA banner at the top and put the licence badge back to the old spelling. The readme/ fragments aren't touched so the bot regenerates both on merge anyway.

nimarosa (changes requested):
I approved this too fast last month, sorry about that. Went back over it properly and there are a few things I want sorted first, the unique constraint on (contract_id, advantage_template_id) being the one that actually matters. The bound fix and the digits work are good. Happy to /ocabot merge minor once these are in, it's more than a patch given the new constraints.

PR #294, opened 2026-09-04
[19.0][MIG] payroll_account: Migration to 19.0

amincheloh:
Migration of `payroll_account` to 19.0.

Part of #216. Depends on `payroll` 19.0.1.0.0, already present in the 19.0 branch.

### Contracts are now employee versions (`hr.contract` → `hr.version`)

- `analytic_account_id` and `journal_id` are declared on `hr.version`.
- Added a `_get_whitelist_fields_from_template()` override so both fields are
  copied when a contract template is applied to an employee — without it, core
  only propagates its own whitelist and the salary journal / analytic account of
  a template would be silently dropped.
- `hr_contract_view_form` now inherits `payroll.hr_contract_view_form` (the
  contract template form) and adds an *Accounting* group after the contract
  group.
- New `hr_employee_view_form` inheritance exposing both fields in the employee
  *Payroll* tab, right after `struct_id`, since contract data is edited from the
  employee form in 19.0.

### Partner resolution on payslip lines

- `hr.employee.bank_account_id` no longer exists; `_get_partner_id()` now falls
  back to `primary_bank_account_id.partner_id` when the employee has no work
  contact.

### Salary rules

- Removed the `[('deprecated', '=', False)]` domain from `account_debit` and
  `account_credit`: `account.account.deprecated` no longer exists in 19.0.

### Misc

- `_()` → `self.env._()` with named interpolation parameters.
- Dropped the obsolete `18.0.1.0.2` migration scripts.
- Tests adapted to the version model (the version created with the employee is
  completed instead of creating an `hr.contract`, `gender` → `sex`,
  `bank_account_id` → `bank_account_ids`), plus a new test covering the bank
  account fallback in the partner logic.

## Notes for reviewers

The branch replays the module history from 18.0; administrative commits
(Weblate, `[BOT] post-merge updates`, `.pot` and README regenerations) have been
squashed into their preceding functional commit as suggested by the migration
wiki.


amincheloh:
@nimarosa please review PR, thanks.

prayagrk (approved):
Code review, LGTM!

PR #298, opened 2026-09-17
[18.0][FIX] payroll: salary rule name column truncated in lists

nimarosa:
In the salary rule lists every column gets the same share of the available width, so `name` — the only column a user reads to tell two rules apart — is truncated with an ellipsis while `code` and `register_id` keep space they do not need. It is most visible in the rule list embedded in the salary structure form, which is where rules are actually browsed.

* `code` and `category_id` are pinned to a fixed width (`width="Npx"` sets both the min and max width of a list column, so the column stops growing) and `name` is left unconstrained, taking whatever is left.
* `register_id` is empty on most rules and becomes an optional column.
* `sequence` becomes a handle in both rule lists — the order of the rules in a structure is exactly what one edits there; in the structure form's list it was `column_invisible`, i.e. present but unusable.
* The salary structure **list** rendered `rule_ids` as a comma separated blob of every rule name, which pushes `name`, `code` and `company_id` out of view on any structure with more than a couple of rules. It is now an optional `many2many_tags` column.
* The main salary rule list hid `category_id` with `column_invisible="True"`, although the search view offers grouping by category; it is an ordinary optional column now.
* The salary structure form gets the usual `oe_title` heading for `name`, like the salary rule form next to it.

Touches `hr_salary_rule_view_tree_children` as well, which is also cleaned up in the sibling PR "[FIX] payroll: dead and duplicate list columns" (duplicate `sequence` field) — the two conflict textually, not semantically.

@OCA/payroll-maintainers


pedrobaeza:
@nimarosa, beware with the AI...

nimarosa:
@pedrobaeza Yes, it was not meant to be in this repo. I closed all of them.

pedrobaeza:
Well, they may be correct for OCA, but check https://github.com/pypa/pip/blob/main/AI_POLICY.md

